PR #1, opened 2018-03-07
Update README.md

fegol:
fixed command block position


seik:
Hi, thank you for your fix!

PR #2, opened 2018-04-11
Inline keyboard markup feature

SavinMike:
[feature][inlineButtons] Supported inline keyboard markup (https://core.telegram.org/bots#inline-keyboards-and-on-the-fly-updating);
[feature][payload] Added extension to get payload list from start command;
[feature][command] Added extension to start command;

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -24,8 +24,6 @@ telegram/local.properties
 # gradle/wrapper/gradle-wrapper.properties
 ### Java template
 # Compiled class file
-*.class
-
 # Log file
 *.log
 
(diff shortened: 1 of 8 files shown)


seik:
Hi @SavinMike! Thank you for your work. It looks really good.

I have done some changes before merging it (sorry 🤓). I've removed the ```start``` extended function, it's a really good idea but since the ```command``` method exist I feel that it's not really a necessity. And If someone wants to do something like that they can write that function on their own project (since extended functions are awesome 😜). Rest of the changes are styling and minor stuff.

Again, thank you for the code <3. Let's merge this!

PR #19, opened 2018-07-26
Add Filters

hangyas:
Adds Filters, see [sample](https://github.com/hangyas/kotlin-telegram-bot/blob/0d243bd4199032e0892a219975fe56366fc6ab5a/samples/filters/src/main/kotlin/me/ivmg/filters/Main.kt) about how it works. It's more like a proposal, there might be better ways to implement this and use cases I forgot to consider.

A few things to note:
 - `and` doesn't take precedence over `or`, ~maybe we should wait for https://github.com/JetBrains/kotlin/pull/1769 instead of this~
 - an other option is to use `+` and `*`
 - some builtin filters could be implemented later, like `message contains url`, etc.

### DIFF
--- a/telegram/src/main/kotlin/me/ivmg/telegram/Bot.kt
+++ b/telegram/src/main/kotlin/me/ivmg/telegram/Bot.kt
@@ -98,7 +98,7 @@ class Bot private constructor(
         parseMode: ParseMode? = null,
         disableWebPagePreview: Boolean? = null,
         disableNotification: Boolean? = null,
-        replyToMessageId: Int? = null,
+        replyToMessageId: Long? = null,
         replyMarkup: ReplyMarkup? = null
     ) = apiClient.sendMessage(
         chatId,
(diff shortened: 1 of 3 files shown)


Redrield:
I wouldn't hold your breath for Jetbrains/Kotlin#1769 to be merged. In the associated YouTrack issue folks at Jetbrains expressed some resistance towards bitwise operators, and it's already missed the Kotlin 1.3 deadline

Redrield:
The reasoning that Andrey gave for closing it seems to imply that if I follow due process it may get reopened in the future. In the mean time however infix functions are your only choice unfortunately

seik:
Hi! Right now I'm on vacation, I come back in a week or so, will see then what we do with this ;).

It looks really good, the only thing that I have in mind is if there is some other way to achieve this and if those ways are easier in terms of using the filters (not internally). Overall I like the idea and implementation.

PR #20, opened 2018-08-05
Fixes

hangyas:
Fixing a few issues


seik:
@hangyas Thanks for the fixes!

PR #23, opened 2018-09-14
Remove unused entities

SavinMike:
Payment entities was moved to payment folder. I was removed old files


seik:
Hey! Those files are [used](https://github.com/seik/kotlin-telegram-bot/blob/master/telegram/src/main/kotlin/me/ivmg/telegram/entities/Message.kt#L47) 🤔

Thanks for the PR anyways 😄

PR #25, opened 2018-09-17
handle channels updates

skel2007:
current handlers can't handle messages from channels bot is subscribed to


telegram/src/main/kotlin/me/ivmg/telegram/dispatcher/handlers/ChannelTextHandler.kt
```diff
@@ -0,0 +1,13 @@
+package me.ivmg.telegram.dispatcher.handlers
+
+import me.ivmg.telegram.HandleUpdate
+import me.ivmg.telegram.entities.Update
+
+class ChannelTextHandler(handleUpdate: HandleUpdate) : Handler(handleUpdate) {
```
seik: First of all thank you for your PR. 

`channel_post` contains a message, which can be a photo, text or any other thing as the telegram docs say: 

> Optional. New incoming channel post of any kind — text, photo, sticker, etc.

So `ChannelTextHandler` could be changed to `ChannelHandler`

PR #29, opened 2019-01-23
Changed type from Boolean? to Long?

Flassie:
Changed `migrateToChatId` and `migrateFromChatId` that fixes https://github.com/seik/kotlin-telegram-bot/issues/28 issue

### DIFF
--- a/samples/dispatcher/src/main/kotlin/me/ivmg/dispatcher/Main.kt
+++ b/samples/dispatcher/src/main/kotlin/me/ivmg/dispatcher/Main.kt
@@ -4,6 +4,7 @@ import me.ivmg.telegram.HandleUpdate
 import me.ivmg.telegram.bot
 import me.ivmg.telegram.dispatch
 import me.ivmg.telegram.dispatcher.callbackQuery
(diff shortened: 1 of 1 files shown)


telegram/src/main/kotlin/me/ivmg/telegram/Bot.kt
```diff
@@ -72,7 +72,7 @@ class Bot private constructor(
             false, null -> {
                 val errorMessage: String = when {
                     error?.message != null -> error.message!!
-                    response.errorBody() != null -> response.errorBody().toString()
+                    response?.errorBody() != null -> response?.errorBody().toString()!!
```
seik: Can this line be:

```
response?.errorBody() != null -> response.errorBody().toString()
```

Since we have already checked out `response?.errorBody()` is not null, (I think) there is no way `response` is null, right?

seik:
Hi, thanks for your pr! I have added a comment on one of the changed files 😄

seik:
Merged this changes manually since I fucked up with the merge. Thanks for your contribution!

PR #38, opened 2019-10-02
Add method to stop bot itself

### LINKED ISSUE #30: Allow to stop bot instance
For now bot can be started but can not be stopped. In my use-case I need to change proxy value at runtime or disable bot at all but it's not possible now. 

So I think Closeable interface could be implemented and after close() called executor and dispatcher should be closed.

P.s. code is so gorgeous!

JcMinarro:
Create a method to be able to stop the bot.
The same bot instance could be started after it is stoped.
It fixes #30

### DIFF
--- a/telegram/src/main/kotlin/me/ivmg/telegram/Updater.kt
+++ b/telegram/src/main/kotlin/me/ivmg/telegram/Updater.kt
@@ -8,17 +8,19 @@ import me.ivmg.telegram.entities.Update
 class Updater {
     private val executor: Executor = Executors.newCachedThreadPool()
     private var lastUpdateId = 0L
+    private var stopped = false
 
     lateinit var bot: Bot
     val dispatcher = Dispatcher()
 
     fun startPolling() {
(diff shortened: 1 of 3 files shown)


telegram/src/main/kotlin/me/ivmg/telegram/Updater.kt
```diff
@@ -36,4 +38,9 @@ class Updater {
             lastUpdateId = lastUpdate.updateId + 1
         }
     }
+
+    internal fun updaterStopPolling() {
```
seik: I think this should be called `stopPolling`.

I understand you used `updaterStopPolling` because of `updaterStartPolling`, but in that case was just because we have the public method used as a "launcher" and the private one.

JcMinarro:
Ok @seik I have renamed the method ;)

seik:
Looking good. LGTM!

Will release a new version with as soon as I get one or two functionalities more to pack.

Issue #33, opened 2019-05-25
Allow inline query handling


BjoernPetersen:
There is an `InlineQuery` entity, but otherwise the whole inline query feature is ignored right now.

seik:
This has been implemented in #42

PR #42, opened 2019-10-26
Support for bot's inline mode (#33)

vjgarciag96:
As requested per #33, in this PR you'll see the changes needed to fully support telegram bot's inline mode as specified in telegram bot api documentation (https://core.telegram.org/bots/api#inline-mode). The main changes you'll see here are the next:

- New entities needed to support API operations related with inline mode. You can check them at `InlineQueryResult.kt` and `InlineMessageContent.kt` files.

- `anwerInlineRequest` API operation. You can check the changes in `ApiClient.kt` and `Bot.kt`. Moreover, other change has been needed to support this API operation; a custom serializer for `InlineQueryResult` entity. You can check it at `InlineQueryResultAdapter.kt`

- New handler for inline queries. You can check the changes at `Callbacks.kt`, `Dispatcher.kt` and `InlineQueryHandler.kt` files.

- Finally, an inline query sample has been added to showcase how to use it.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -5,7 +5,7 @@ buildscript {
 }
 
 plugins {
-    id "org.jetbrains.kotlin.jvm" version "1.2.50"
+    id "org.jetbrains.kotlin.jvm" version "1.3.50"
 }
 
 allprojects {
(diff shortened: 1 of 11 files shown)


build.gradle
```diff
@@ -5,7 +5,7 @@ buildscript {
 }
 
 plugins {
-    id "org.jetbrains.kotlin.jvm" version "1.2.50"
+    id "org.jetbrains.kotlin.jvm" version "1.3.50"
```
vjgarciag96: Needed to make contracts work (currently using them with `checkNotNull` function in order to let the compiler to smart cast the nullable type to the non nullable one in some parts of the code)

seik:
This looks really really good, LGTM!

Issue #34, opened 2019-06-14
Add file dispatcher


MatthiasReumann:
It would be nice to have a dispatcher like the location one for files.

PR #43, opened 2019-10-26
Dispatchers for files (#34)

vjgarciag96:
As requested per #34, in this PR you'll see the changes needed to have dispatchers like the location one for the different  types of files. 

Although this could be achieved with message dispatcher plus filters I think is good to also offer these new ones to allow users having the `sticker{...` syntax instead of `message(Filters.Photo)` and also to allow them working directly with the concrete file type instead of with messages (using message dispatcher plus filters we will have to still check manually if the media type we expect is not null. With the new approach we could work directly with the non nullable media type)


build.gradle
```diff
@@ -5,7 +5,7 @@ buildscript {
 }
 
 plugins {
-    id "org.jetbrains.kotlin.jvm" version "1.2.50"
+    id "org.jetbrains.kotlin.jvm" version "1.3.50"
```
vjgarciag96: Needed to make contracts work (currently using them with `checkNotNull` function in order to let the compiler to smart cast the nullable type to the non nullable one in some parts of the code)

alex1815:
Hey @vjgarciag96 ,
Nice PR. I'm looking for audio support for bot. Can I help you with this PR? Or should we ping @seik  for review of this one and wait?

vjgarciag96:
Hi @alex1815 !! In this PR you also have the support for audio implemented, you can check it at https://github.com/seik/kotlin-telegram-bot/pull/43/commits/456434816623abbaa32ef37413c24876e0fab12d#diff-77a6ef1e09404cde07316a574ceb7643R87 . Or are you making reference to a different thing?

alex1815:
I mean have you completed all work with this PR and we're waiting for review form @seik side before merge in master? Would be nice to get these changes

seik:
@vjgarciag96 @alex1815 all merged!

Issue #37, opened 2019-09-15
Downloading files


sssemil:
In the current version it is not possible to dowload files, like photos or anything else.

Impement the 'file' call, than can use the File returned by the [getFile](https://core.telegram.org/bots/api#getfile) call.

No alternatives I can see, except for dowloading it manually, in your own code.

Just implement dowloading based on File object with `https://api.telegram.org/file/bot<token>/<file_path>` link, as described in the API docs,

rundoom:
Seems like this issue can be closed by 9889665109fd9229eaed57ac0d84b36816dc4b87

PR #46, opened 2019-12-29
Downloading files and sending documents from ByteArray

rundoom:
File downloading from issue
https://github.com/kotlin-telegram-bot/kotlin-telegram-bot/issues/37

Sending documents with ByteArray type(useful if you have no actual filesystem file)

### DIFF
--- a/telegram/src/main/kotlin/me/ivmg/telegram/network/ApiClient.kt
+++ b/telegram/src/main/kotlin/me/ivmg/telegram/network/ApiClient.kt
@@ -48,12 +48,13 @@ private fun convertFile(name: String, file: SystemFile, mimeType: String? = null
 private fun convertBytes(
     name: String,
     bytes: ByteArray,
-    mimeType: String? = null
+    mimeType: String? = null,
+    filename: String? = null
 ): MultipartBody.Part {
     val mediaType = mimeType?.let { MediaType.parse(it) }
     val requestBody = RequestBody.create(mediaType, bytes)
 
-    return MultipartBody.Part.createFormData(name, name, requestBody)
+    return MultipartBody.Part.createFormData(name, filename ?: "untitled", requestBody)
 }
 
 class ApiClient(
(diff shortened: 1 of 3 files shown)


telegram/src/main/kotlin/me/ivmg/telegram/network/ApiClient.kt
```diff
@@ -51,17 +52,18 @@ private fun convertFile(name: String, file: SystemFile, mimeType: String? = null
...
     bytes: ByteArray,
-    mimeType: String? = null
+    mimeType: String? = null,
+    filename: String? = null
 ): MultipartBody.Part {
     val mediaType = mimeType?.let { MediaType.parse(it) }
     val requestBody = RequestBody.create(mediaType, bytes)
 
-    return MultipartBody.Part.createFormData(name, name, requestBody)
+    return MultipartBody.Part.createFormData(name, filename ?: "untitled", requestBody)
```
vjgarciag96: I would make `filename` mandatory in the new `sendDocument` method added to support `ByteArray` and wouldn't add "untitled" as default value for it on `convertBytes` method. I think this could be confusing for someone using the mentioned method and that's better to make mandatory indicating the filename in that case. If user wants to have "untitled" as title they still can, but will be totally under their responsibility. What do you think about it?

telegram/src/main/kotlin/me/ivmg/telegram/network/ApiClient.kt
```diff
@@ -258,6 +260,28 @@ class ApiClient(
...
+
+        return service.sendDocument(
+            convertString(chatId.toString()),
+            convertBytes("document", fileBytes, null, fileName ?: caption),
+            if (caption != null) convertString(caption) else null,
+            if (parseMode != null) convertString(parseMode) else null,
+            if (disableNotification != null) convertString(disableNotification.toString()) else null,
+            if (replyToMessageId != null) convertString(replyToMessageId.toString()) else null,
+            if (replyMarkup != null) convertJson(replyMarkup.toString()) else null
+        )
```
vjgarciag96: I would probably change `convertXXX` functions to make them extension functions of `String`, `ByteArray`, etc... to take advantage of Kotlin features. In this case, if `convertString` would have been an extension function the code involving nullables could be easier to read:

```
return service.sendDocument(
    chatId.toString().convertString(),
    ....,
    caption?.convertString(),
    parseMode?.convertString(),
    ...
```

However, that's probably out of scope of this PR, so you can keep it as it is right now and we can consider adding the mentioned functions in a future task.

telegram/src/main/kotlin/me/ivmg/telegram/network/ApiClient.kt
rundoom: Hello! @vjgarciag96 I think mandatory `filename` parameter might be more clear for usage. Do you want me to refactor my code or you will do it by your self?

telegram/src/main/kotlin/me/ivmg/telegram/network/ApiClient.kt
rundoom: That's fine, but it also can be done by using `?.let{}` like this `caption?.let { convertString(it) }`

telegram/src/main/kotlin/me/ivmg/telegram/network/ApiClient.kt
vjgarciag96: As you want, I can do the changes but you'll probably have to wait to the weekend. If you're ok with that, I can take care of it.

telegram/src/main/kotlin/me/ivmg/telegram/network/ApiClient.kt
vjgarciag96: Sure, I prefer the extension function option because is more readable IMHO but it's probably a matter of taste. Anyways, these changes are out of the scope of the PR you did, maybe we can discuss them for a future task.

PR #50, opened 2020-02-09
Downgraded logging-interceptor

red-avtovo:
Your retrofit dependency is using `okhttp` library version 4.3.1
Since your okhttp logging interceptor is older it throws an error each time I start the bot:
```
Exception in thread "pool-1-thread-2" java.lang.NoSuchMethodError: okhttp3.internal.platform.Platform.log(ILjava/lang/String;Ljava/lang/Throwable;)V
	at okhttp3.logging.HttpLoggingInterceptor$Logger$1.log(HttpLoggingInterceptor.java:110)
	at okhttp3.logging.HttpLoggingInterceptor.intercept(HttpLoggingInterceptor.java:160)
	at okhttp3.internal.http.RealInterceptorChain.proceed(RealInterceptorChain.kt:112)
	at okhttp3.internal.http.RealInterceptorChain.proceed(RealInterceptorChain.kt:87)
	at okhttp3.RealCall.getResponseWithInterceptorChain(RealCall.kt:194)
	at okhttp3.RealCall.execute(RealCall.kt:67)
	at retrofit2.OkHttpCall.execute(OkHttpCall.java:180)
	at me.ivmg.telegram.network.UtilsKt.call(Utils.kt:8)
	at me.ivmg.telegram.Bot.getUpdates(Bot.kt:70)
	at me.ivmg.telegram.Updater.updaterStartPolling(Updater.kt:22)
	at me.ivmg.telegram.Updater.access$updaterStartPolling(Updater.kt:8)
	at me.ivmg.telegram.Updater$startPolling$2.run(Updater.kt:17)
	at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1142)
	at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:617)
	at java.lang.Thread.run(Thread.java:745)
```

I don't know why you don't have a similar issue, but I fixed in my project by bumping up logging-interceptor version

### DIFF
--- a/telegram/build.gradle
+++ b/telegram/build.gradle
@@ -6,7 +6,7 @@ buildscript {
 
     ext.retrofit_version = '2.3.0'
     ext.junit_version = "5.5.1"
-    ext.logging_interceptor_version = '4.3.1'
+    ext.logging_interceptor_version = '3.8.0'
 
     repositories {
         mavenCentral()
--- a/telegram/build.gradle
+++ b/telegram/build.gradle
@@ -6,7 +6,7 @@ buildscript {
 
     ext.retrofit_version = '2.3.0'
     ext.junit_version = "5.5.1"
-    ext.logging_interceptor_version = '3.9.1'
+    ext.logging_interceptor_version = '4.3.1'
 
     repositories {
         mavenCentral()


vjgarciag96:
Hi @red-avtovo!! I've been checking the issue you commented and the conclusions I've extracted are the next:

* Currently in `master` branch, the dependency tree is the next: 

master

As you can see, retrofit 2.3.0 version depends on okhttp 3.8.0 and the okhttp interceptor depends on okhttp 3.9.1. However, gradle resolves the okhttp dependency conflict by updating the retrofit's okhttp version to 3.9.1.

* In your branch, the dependency tree changes to the next:

PR

And gradle resolves the conflict updating retrofit's okhttp version to the greater one again.

Therefore I'd say the problem you were having was caused by the way gradle was resolving the dependencies, dunno exactly why. Maybe we can get a clue if you run the `./gradlew telegram:dependencies` command on `master` branch.

Anyways, the only way to fix this "problem" would be changing the okhttp interceptor version to the one our retrofit's version is using, which is currently 3.8.0. What do you think about this?

Thanks for reporting and sorry for the delay in the response.

red-avtovo:
Make sense indeed. 
It is just unusual to downgrade the library instead of aligning every other one to use a newer set of libraries. I will update the PR now with interceptor version `3.8.0` since it is using okhttp version `3.8.0` as you can see on a screenshot below from maven central

https://mvnrepository.com/artifact/com.squareup.okhttp3/logging-interceptor/3.8.0

image

vjgarciag96:
I'm going to merge it, thanks for the PR!!

PR #51, opened 2020-04-05
feat: introduced getMyCommands and setMyCommands methods

vicmosin:


telegram/src/main/kotlin/me/ivmg/telegram/entities/BotCommand.kt
```diff
@@ -0,0 +1,9 @@
+package me.ivmg.telegram.entities
+
+import com.google.gson.annotations.SerializedName as Name
+import me.ivmg.telegram.entities.stickers.ChatPermissions
```
vjgarciag96, line 4: Really minor: unused imports here.

If you run the `./gradlew ktlintFormat` command, `ktlint` will take care of this things for you.

Issue #44, opened 2019-11-04
Webhook


elihun:
Hello! Is there any way to use webhooks for update handling?

seik:
Hi @ilya-gr, right now there is no option for webhooks. I can't say I'm going to implement it since it's something I'm not going to use by myself.

If someone wants to give a take on it the PRs are open! I'm going to leave this issue ope as a reference.

ilya-gr:
Hi @seik, thank you for the answer!

vjgarciag96:
I'm going to start working on this, stay tuned!

vicmosin:
@vjgarciag96 Is there any update regarding the issue?

vjgarciag96:
Hi @vicmosin!! I have a branch with a working version. However, I wanted to test that all the certificates stuff is working in a real environment, but I had no time to finish checking that. Do you want to contribute on this?. If so, I could upload the work in progress version of my branch to let you check how it's going and to also let you contributing to it if you want.

vjgarciag96:
Here you can check the advances of my branch -> https://github.com/vjgarciag96/kotlin-telegram-bot/commit/7cb55145f2002e01e1d46a211adab9b7edd75670

If you want to contribute on it, please let me know (cc/ @vicmosin)

vicmosin:
@vjgarciag96 Could you please put all the changes into your branch and describe a bit what's still missing? I think I could check it

vjgarciag96:
Sure, the branch is now updated with the latest changes!

Regarding the changes of the branch...

I have basically added the three API operations needed to work with webhooks:

- `setWebhook`. This is the one that has more complexity. Well, the only "complex" point is that the certificate file can be sent in different ways (using a file id, an url of an already uploaded file or as a `multipart/form-data` chunk)
- `getWebhookInfo`
- `deleteWebhook`

As the size of the current API client was getting too big, I created another class to handle the`setWebhook` operation, which is a new collaborator of the current API client. The current API client just delegates the calls to `setWebhook` to it. Maybe we could find an abstraction for operations involving files and avoid this new class, but it's probably out of scope of this PR.

In addition, the `Bot` class has been slightly updated in order to support updates from either the webhook or the `getUpdates` method (previous implementation was not ready for that).

Finally, a new module has been added with a sample bot using webhooks that is run within a `Netty` server with `ktor` framework.

[...]

PR #52, opened 2020-04-21
(#44) Support webhooks to receive bot's updates

vjgarciag96:
In this PR you'll see the changes needed to add support for webhooks as a new method to get bot updates. Telegram offers three main operations related with webhooks:

* `setWebhook`: to set up a new webhook
* `getWebhookInfo`: to get information about the currently set webhook
* `deleteWebhook`: to delete the currently set webhook

We want to add support in our library for these three operations. Not only that, but we also want to provide a friendly and readable way to set up a webhook using our bot building dsl.

The main changes you'll find in the PR are the next:

* The new API operations. You can check the changes on `TelegramFile`, `WebhookInfo`, `ApiClient`, `ApiConstants`, `ApiService`, `MediaTypeConstants` and `MultipartExtensions` files. The `deleteWebhook` and `getWebhookInfo` operations are quite simple, we just need to send an HTTP GET with the correct operation name and we're done. The `setWebhook` is a bit harder. This operation accepts several parameters. One of them is a certificate file that we could need to send to Telegram to let their servers trust in a self-signed certificate, for example. When sending files, Telegram accepts three ways:

    1. Sending the file id of an already uploaded to a Telegram server file.
    2. Sending a file url.
    3. Sending the file as a part of a multipart request.

Therefore we need to provide methods to set the webhook by using any of these ways to upload files. The first two are simple, just a usual HTTP POST with the correspondent body. For the last one we need to send a multipart request and also transform to multipart parts all the body params. To do so, we've created the `MultipartExtensions` file where we can find extension functions to transform typical java types to multipart parts.

* Updates on `Bot` class to allow it processing updates from other sources more than the polling component. Moreover, two new methods `startWebhook` and `stopWebhook` have been added to it to allow the user starting the webhook configured on bot set up and stopping it when needed. The `startWebhook` method takes the webhook config added to the bot and calls the `setWebhook` API operation with it. If the webhook is successfully set, the updater starts checking updates and the method returns true. Otherwise, the updater will do nothing and the method will return false. The `stopWebhook` method calls the `deleteWebhook` API operation and tells the updater to stop checking updates. If the `deleteWebhook` operation is successful it returns true and otherwise false

* The bot building dsl has been extended to include a new configuration function, `webhook`. By using this configuration function you can set your webhook with a really nice syntax. Then, to start listening updates, you just need to call the bot's `startWebhook` method (kind of webhook's equivalent of the `startPolling` method used to start the polling). You can check the changes needed to add this syntax in `Bot` and `WebhookConfig` files.
```
bot {
    token = ANY_TOKEN
    webhook {
        url = ANY_URL
    }
}.startWebhook()
```

* Tests. A few tests have been added for the code of the new bot methods `startWebhook` and `stopWebhook`, for the different `setWebhook` alternative implementations and also a new test has been added for the `Dispatcher`.

* Documentation updates. The main README has been updated and a new documentation page has been added with details about the different possibilities of getting updates from Telegram's bots and how you can implement them by using our library.

* A new example showcasing the usage of webhooks has been added. It's just a simple bot that has a `/hello` command that simply answers with a "Hey Bruh!" text. But the interesting part of the sample is that it sets a webhook and the updates are received through a web server. For the sample I've used the ktor framework and an embedded Netty server just for the sake of simplicity and readability, but library users can use whatever server configuration and framework they want.


docs/gettingUpdates.md
```diff
@@ -0,0 +1,98 @@
...
+    // bot implementation
+}.startPolling()
+```
+
+For a more detailed example you can check the `dispatcher` and `echo` samples you can find in the `samples` folder of this project.
+
+
+## Setting up a webhook
+
+Telegram offers a way to receive updates through a webhook. You just have to indicate to Telegram the url where you want to receive the updates and Telegram will send an HTTP POST with updates every time updates are received for a given bot.
```
manolovn, line 35: I think Telegram sends HTTPS POST (https://core.telegram.org/bots/api#setwebhook)

samples/webhook/src/main/kotlin/me/ivmg/webhook/CertificateUtils.kt
```diff
@@ -0,0 +1,33 @@
+package me.ivmg.webhook
+
+import java.io.File
+import java.io.FileInputStream
+import java.security.KeyStore
+
+object CertificateUtils {
+    val privateKeyPassword = "YOUR_PRIVATE_KEY_PASSWORD".toCharArray()
```
manolovn, line 8: I'm not sure if this "toCharArray" is expected; what's the motivation behind?
maybe this password could be constants and the conversion to char array done lazily when needed

vjgarciag96: The motivation is just that the places where we use this password need a `CharArray`. And as we're currently using it in three different places, instead of having a constant `String` and transforming it to `CharArray` every time we need to use it I prefer transforming it to `CharArray` just one time and using it everywhere.

docs/gettingUpdates.md
vjgarciag96: Yes, it's HTTP**S**. I'll update the doc to make it clearer.

Issue #45, opened 2019-12-29
`sendMediaGroup` throws retrofit2 exception


bash-spbu:
I get the following error when invoke `Bot#sendMediaGroup` method :
```traceback
Exception in thread "pool-1-thread-1" java.lang.IllegalArgumentException: Parameter type must not include a type variable or wildcard: java.util.List<? extends me.ivmg.telegram.entities.inputmedia.InputMedia> (parameter #2)
    for method ApiService.sendMediaGroup
	at retrofit2.ServiceMethod$Builder.methodError(ServiceMethod.java:752)
	at retrofit2.ServiceMethod$Builder.methodError(ServiceMethod.java:743)
	at retrofit2.ServiceMethod$Builder.parameterError(ServiceMethod.java:761)
	at retrofit2.ServiceMethod$Builder.build(ServiceMethod.java:195)
	at retrofit2.Retrofit.loadServiceMethod(Retrofit.java:170)
	at retrofit2.Retrofit$1.invoke(Retrofit.java:147)
	at com.sun.proxy.$Proxy0.sendMediaGroup(Unknown Source)
	at me.ivmg.telegram.network.ApiClient.sendMediaGroup(ApiClient.kt:471)
	at me.ivmg.telegram.Bot.sendMediaGroup(Bot.kt:436)
	at me.ivmg.telegram.Bot.sendMediaGroup$default(Bot.kt:435)
```

In my opinion problem is here:  
https://github.com/kotlin-telegram-bot/kotlin-telegram-bot/blob/ab798c96a3a6960a61cc9112842af636b1344543/telegram/src/main/kotlin/me/ivmg/telegram/network/ApiService.kt#L256

It is because Kotlin's List is covariant and to support it Kotlin's compiler generates Java's wildcards. So indeed it's something like 
```java
@Field("media") List<? extends InputMedia> media,
```

As I think the *possible fixes* for it is either to add `@JvmSuppressWildcards` annotation to this parameter
```kotlin
@Field("media") media: List<@JvmSuppressWildcards InputMedia>,
```

either try to use `Array<InputMedia>` or `MutableList<InputMedia>` instead of `List<InputMedia>`.

*Additional info:*
- https://kotlinlang.org/docs/reference/java-to-kotlin-interop.html#variant-generics
- https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.jvm/-jvm-suppress-wildcards/index.html
[...]

PR #60, opened 2020-05-02
sendMediaGroup API operation

vjgarciag96:
This PR fixes the problems reported on #45

The `sendMediaGroup` operation will allow to send a group of media (from 2 to 10 photos and/or videos) as an album to any chat

The media elements composing the group can be sent as file ids, file urls or files (with the possibility of combining all these types of files)

There's an `InputMedia` model defined by Telegram that represents any type of media that can be sent to a chat. This `InputMedia` is not limited to only photos or videos but
also other media types like audio, documents, etc... So, to represent the `InputMedia` model we have a sealed class, and for every concretion (photo, video, etc..) we have a class extending it. In this case, as the type of media that can be sent as media groups is limited to photos and videos, a new interface called `GroupableMedia` has been created to mark all the media types that can be sent as a group and the `InputMediaPhoto` and `InputMediaVideo` classes have been marked with it.

A few tests have been added for the `MediaGroup` class, to check that its validation is working as expected and for the new API operation, where we have added some integration tests where we check that the multipart request body that we send to Telegram's servers is correct.

Also, an example of the usage of this operation has been added to the dispatcher sample


telegram/src/main/kotlin/com/github/kotlintelegrambot/network/serialization/GsonFactory.kt
```diff
@@ -0,0 +1,24 @@
...
+
+    fun createForApiClient(): Gson = GsonBuilder()
+        .registerTypeAdapter(InlineQueryResult::class.java, InlineQueryResultAdapter())
+        .create()
+
+    fun createForMultipartBodyFactory(): Gson = GsonBuilder()
+        .registerTypeAdapter(TelegramFile.ByFile::class.java, TelegramFileAdapter())
+        .registerTypeAdapter(TelegramFile::class.java, TelegramFileAdapter())
+        .registerTypeAdapter(GroupableMedia::class.java, GroupableMediaAdapter(InputMediaAdapter()))
+        .create()
```
vjgarciag96, line 23: I've created different factory methods here to create Gson instances with different configurations to only have the adapters needed in each component and not having all of them in all the instances

telegram/src/main/kotlin/com/github/kotlintelegrambot/network/serialization/adapter/TelegramFileAdapter.kt
```diff
@@ -0,0 +1,18 @@
...
+import com.google.gson.JsonSerializationContext
+import com.google.gson.JsonSerializer
+import java.lang.reflect.Type
+
+class TelegramFileAdapter : JsonSerializer<TelegramFile> {
+    override fun serialize(src: TelegramFile, typeOfSrc: Type, context: JsonSerializationContext): JsonElement = when (src) {
+        is ByFileId -> context.serialize(src.fileId, String::class.java)
+        is ByUrl -> context.serialize(src.url, String::class.java)
+        is ByFile -> context.serialize("attach://${src.file.name}")
+    }
```
vjgarciag96, line 17: This is the way the different types of file must be serialized for the `sendMediaGroup` operation

`attach://filename` indicates that a file with that name is being sent as part of the multipart request body (it's the approach Telegram has stablished for this case)

telegram/src/main/kotlin/com/github/kotlintelegrambot/network/multipart/MultipartExtensions.kt
```diff
@@ -0,0 +1,18 @@
+package com.github.kotlintelegrambot.network.multipart
```
vjgarciag96, line 1: Just moved

telegram/src/main/kotlin/com/github/kotlintelegrambot/entities/inputmedia/InputMedia.kt
```diff
@@ -1,7 +1,112 @@
...
+    abstract val parseMode: String?
+}
+
+/**
+ * Interface to mark all the media types that can be sent within a group of media for
+ * operations like `sendMediaGroup`.
+ */
+interface GroupableMedia
+
+class MediaGroup private constructor(val medias: Array<out GroupableMedia>) {
```
manolovn, line 23: a Set instead of Array could helps here?

vjgarciag96: hmm... the point is that you can send the same media two times, I guess that's not the usual use case, but using a set wouldn't be accurate enough for this reason. 

Also, using an `Array` lets us taking advantage of `varargs` in an easier way

PR #71, opened 2020-05-14
Feature/rename params for consistency

kissedcode:


vjgarciag96:
Hi @kissedcode! 

First of all, thanks for the contribution.

Regarding the changes, I'd prefer to not integrate them because of the next reasons:
* The `sendXXX` (photo, audio, document, etc...) methods offered by Telegram Bot API allow sending the correspondent media in three different ways; by sending the file id, the file url or by sending the file itself as a part of a multipart request. In those methods, we currently support sending the file as file id or as file url. Therefore, using file id as parameter name won't be correct at all because it can be either a file id or a file url.
* To keep consistency with Telegram Bot API. We use the current names they're using in the API methods.
* To avoid breaking changes in the library API. Adding these changes will force to all users that update the library to the release version that which includes these changes to update all the code where they were referencing the updated code without any direct benefit.

What's your opinion about this?

PR #89, opened 2020-07-08
Catch exceptions in handlers

raid-7:
For now any uncaught exception in an update or error handler kills the dispatching thread. This behaviour is unacceptable because a small mistake in one handler can make the bot unavailable for all users.

Moreover, though without dispatching thread the application actually stops serving, the process remains live and cannot be restarted by an external superviser.

This pull request adds try blocks into Dispatcher to catch errors from handlers and prevent killing of the thread.


vjgarciag96:
Hi @raid-7!! Thank you so much for the contribution! I'll take a look to the PR as soon as possible (probably Friday or Saturday)

Issue #96, opened 2020-07-19
Migrate handlers to lamdas with receivers


vjgarciag96:
In https://github.com/kotlin-telegram-bot/kotlin-telegram-bot/issues/88, we updated command handlers to use a lamda with receiver instead of a lambda in order to provide all the arguments correspondent to the handler in the scope of the lambda, to make them easier and to have a better syntax.

It would be really nice to migrate all the other handlers to the same approach. 

For example, currently we can write a message handler in the next way:

```kotlin
message(Filter.Reply or Filter.Forward) { bot, update ->
    bot.sendMessage(update.message!!.chat.id, text = "someone is replying or forwarding messages ...")
}
```

Our goal would be to have the next:
```kotlin
message(Filter.Reply or Filter.Forward) {
    bot.sendMessage(message.chat.id, text = "someone is replying or forwarding messages ...")
}
```

In this PR you can see how we can implement something like that -> https://github.com/kotlin-telegram-bot/kotlin-telegram-bot/pull/91

vjgarciag96:
Done in https://github.com/kotlin-telegram-bot/kotlin-telegram-bot/pull/103

PR #103, opened 2020-08-09
Handlers to functions with receiver

vjgarciag96:
This is finally ready.

I have migrated all the handlers available in the library to now use functions with receivers. In this way it'll be possible to access to all the data correspondent to the updates in the scope of the handler function without the need of explicitly indicating these data parameters in the handler lambda function.

There is one commit per migrated handler (all the commits that reference the issue https://github.com/kotlin-telegram-bot/kotlin-telegram-bot/issues/96)

Also, there are some clean up commits:

* https://github.com/kotlin-telegram-bot/kotlin-telegram-bot/pull/103/commits/e1978f8e0a343323878028a6bea516a582311fb2 
* https://github.com/kotlin-telegram-bot/kotlin-telegram-bot/pull/103/commits/8d7d45df9b2760b8de799f291efe3cf20db1b069
* https://github.com/kotlin-telegram-bot/kotlin-telegram-bot/pull/103/commits/c9a0b5f6e5ec34db130863ba620c11072270b702

If these changes look good to you, we can finally proceed with this issue https://github.com/kotlin-telegram-bot/kotlin-telegram-bot/issues/93

PR #108, opened 2020-10-02
Handlers to functions with receiver without proxy

reopold:
Hello.
I modified **vjgarciag96** branch handlers-to-functions-with-receiver.
Handlers to functions with receiver #96 can be implemented a little easier, without a proxy.
Please take a look.
Don't kick me too hard, this is my first experience on github and writing in Kotlin.

_P.S.: I didn't even need to add HandlerEnvironment, but I need it for the next PR(chain of steps)._
Thanks.


seik:
Hi @reopold, why did you close this PR?

I currently reviewed #103 and If there is something we can add to that PR or way to improve it I'm open to hear that.

reopold:
Sorry, this is my first experience on github.
I first made PR, and then I saw that I have not all the tests pass. 
I'm trying to fix the tests.
Also, I do not know how better to do this PR. 
I continued **vjgarciag96** branch.  I rewrote his code.
What should I do?
Make a PR to his repository?
Or make a PR here that includes all his commits?

seik:
Since @vjgarciag96 did an incredible job doing that PR I would feel bad merging this. 

Let's do what you are saying, create a PR on @vjgarciag96 repo on the same branch he is doing the PR here, review the changes with him and then after reviewing that we will merge his PR here.

I feel this way is the most fair.

vjgarciag96:
@seik: I'd say these are two different things, so we can easily split them in different PRs.

My PR is about changing our updates handling functions to now implicitly have access to the updates' arguments in the context of the function instead of having to explicitly indicate them (and also bringing better type support for those arguments).
 
However, my PR doesn't include any change to the `Handler` class and doesn't use a different pattern we weren't using before (we use the proxy pattern we'd been using in other handlers)

The PR of @reopold is mostly about reworking how handlers are implemented. The PR includes changes to the `Handler` class to add a new method and also changes all of the implemented handlers to get rid of the proxies we were using before.

IMO we could merge the first PR, and once merged, we can cherrypick the changes proposed here to a new branch checked out from main/master. Then, we could review and discuss the new changes in an easier and clearer way, and the risks would be lower.

What are your thoughts?

reopold:
I agree with @vjgarciag96. 
I will wait for the merge of his branch, and then suggest my PR to improve the implementation.
This will be more consistent.

seik:
Much better approach. Thank you all!

Issue #114, opened 2020-10-21
Abnormal errorMessage returned by getUpdates


reopold:
**Expected behavior**
In case of an error in the process of receiving updates, we would like to receive a response from the server, like:
_{"ok":false,"error_code":409,"description":"Conflict: terminated by other getUpdates request; make sure that only one bot instance is running"}_

**Actual behavior**
Now we get:
_serverResponse = okhttp3.ResponseBody$1@7fe86edf_

**To Reproduce**
Run two instances of the bot.

vjgarciag96:
Fixed in https://github.com/kotlin-telegram-bot/kotlin-telegram-bot/pull/115

PR #115, opened 2020-10-21
Fix abnormal errorMessage (#114)

reopold:
The function **toString()** of errorBody does not return the response text.
It returns something like "okhttp3.ResponseBody$1@7fe86edf"
The function **string()** returns the response text.

I suggest the following algorithm: 
 - if errorBody contains some text, then return it, 
 - else return code and message from server (like "404 Not found").

### DIFF
--- a/telegram/src/main/kotlin/com/github/kotlintelegrambot/Bot.kt
+++ b/telegram/src/main/kotlin/com/github/kotlintelegrambot/Bot.kt
@@ -145,13 +145,29 @@ class Bot private constructor(
                 if (updates?.result != null) return updates.result
             }
             false, null -> {
-                val errorMessage: String = when {
-                    error?.message != null -> error.message!!
-                    response?.errorBody() != null -> response.errorBody().toString()
-                    else -> "There was a problem retrieving updates from Telegram server"
+                fun dispatchableError(message: String) =
+                    listOf<TelegramError>(RetrieveUpdatesError(message))
+
+                val errorMessage = error?.message
+                if (errorMessage != null) {
+                    return dispatchableError(errorMessage)
                 }
 
-                return listOf(RetrieveUpdatesError(errorMessage) as TelegramError)
+                val errorBody = response?.errorBody()?.string()
+                if (errorBody != null && errorBody.isNotBlank()) {
+                    return dispatchableError(errorBody)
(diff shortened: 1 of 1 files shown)


telegram/src/main/kotlin/com/github/kotlintelegrambot/Bot.kt
```diff
@@ -147,7 +147,10 @@ class Bot private constructor(
             false, null -> {
                 val errorMessage: String = when {
                     error?.message != null -> error.message!!
-                    response?.errorBody() != null -> response.errorBody().toString()
+                    (response?.errorBody()?.contentLength() ?: 0L) != 0L ->
+                        response?.errorBody()!!.string()
+                    response?.raw()?.message() != null ->
+                        "${response.raw().code()} ${response.raw().message()}"
```
vjgarciag96, line 153: I was thinking how to make this a bit more readable. What do you think about something like the next?:

```kotlin
            false, null -> {
                fun dispatchableError(message: String) =
                    listOf<TelegramError>(RetrieveUpdatesError(message))
                
                val errorMessage = error?.message
                if (errorMessage != null) {
                    return dispatchableError(errorMessage)
                }

                val errorBody = response?.errorBody()?.string()
                if (errorBody != null && errorBody.isNotEmptyAndNotBlank()) {
                    return dispatchableError(errorBody)
                }

                val rawHttpResponse = response?.raw()
                if (rawHttpResponse != null && rawHttpResponse.message().isNotEmptyAndNotBlank()) {
                    return dispatchableError(
                        "${rawHttpResponse.code()} - ${rawHttpResponse.message()}"
                    )
                }

                return dispatchableError(
                    "There was a problem retrieving updates from Telegram server"
                )
            }
```

vjgarciag96 (commented):
Thank you so much for the PR @reopold! I added a little comment - once resolved I'll be happy to merge this

telegram/src/main/kotlin/com/github/kotlintelegrambot/Bot.kt
reopold: Yes, in this form, readability is much better.

vjgarciag96 (approved):
LGTM!

PR #137, opened 2020-12-12
Add missing `sendVoice` parameters

Flame239:
According to [telegram bot api for sendVoice](https://core.telegram.org/bots/api#sendvoice), it also accepts 3 parameters not presented in current api: `caption`, `parseMode` and `captionEntities`. 

P.S> not sure how I suppose to write tests for it, although I tested it manually.


vjgarciag96 (approved):
LGTM!

Regarding the tests for this, we have kind of integration tests with a mocked server for the API operations. In this way we can test that a given operation in our API implementation is generating the correct HTTP request and is correctly processing the HTTP response that we expect the operation will return. Here's an example of this type of test for the `sendMediaGroup` API operation (https://github.com/kotlin-telegram-bot/kotlin-telegram-bot/blob/master/telegram/src/test/kotlin/com/github/kotlintelegrambot/network/apiclient/SendMediaGroupIT.kt). Sometimes what I do is testing the operation manually with a tool like Postman (https://www.postman.com/ - it could be any other like `curl` for example), collect a sample correct HTTP request and response for the operation, and use that request/response as the "input/output" of my test. Let me know if you want to write the test for that operation, I'm happy to help with that if needed.

Totally out of scope of the PR: I find the `sendVoice` method that receives a string to represent the voice argument a bit confusing because of that argument name . It's currently called `audioId` to represent the file id of a voice file already stored on Telegram servers. That's fine, but a string argument in that operation could also be the url of an .ogg file, and in that case the argument name is totally wrong. I'd change that to `voice` and specify in the documentation that it can be either a file id or a file url.

Flame239:
Alright, I've added test for `sendVoice`

Flame239:
Not sure why test failed 🤔

vjgarciag96:
@Flame239 thanks for adding the test!

That's weird, I remember having a similar problem with a test, but I don't recall how I fixed it. I'll try to take a look to it later to see if I can spot what's going on.

Flame239:
Can we get test report?

vjgarciag96:
That's what I'm trying on https://github.com/kotlin-telegram-bot/kotlin-telegram-bot/pull/138. 

It's not totally working as I expected it to work but I think the current changes can let us to get a report for the test we're interested in. I'll merge it and then we can run the new workflow on this PR to see if we can spot something in the report.

vjgarciag96:
@Flame239 it's now merged. Can you update your branch with the latest changes of master to generate a test report?

vjgarciag96:
Sorry for the noise @Flame239

I was missing one thing in the changes to make them work when the tests actually fail. It should be fixed now and if you update the branch the report should be generated.

vjgarciag96:
Thanks! Now it generated the report, you can find it here https://github.com/kotlin-telegram-bot/kotlin-telegram-bot/suites/1670986396/artifacts/31163316.

That will download folder with an HTML report. If you open the `index.html` file, it won't work. But if you go to the `classes` folder and find the one correspondent to your test, you'll be able to see the report. 

Let me know if you need help with something.

Flame239:
It turns out it adds `Content-[Type: audio/x-vorbis+ogg` to request for some reason

vjgarciag96:
Oh, that's interesting 🤔

I can investigate a bit about why that's happening later if you don't have time to do it.

Flame239:
In `ApiClient.convertFile`  there is some magic code to extract media type. I have debugged it real quick and mediaType is still null on my computer. Looks like it varies between OS'es

vjgarciag96:
I see.... 

I can see a couple of options here.

1. The `convertFile` method tries to infer the content type from a file when we don't indicate it. In the case of the `sendVoice` operation we know that the content type is always has to be `audio/ogg`, so we could indicate it in the `convertFile` call and avoid the inference. That would fix the problem.
2. We could parameterise that part in the expected request we use in the test to make it independent of the inferred part.

I would go with the first one, in this case we know the content type beforehand so it's not needed to try to infer it at all.

Flame239:
Alright, it works now 👌

vjgarciag96 (approved):

vjgarciag96:
Great, thank you so much for the effort! I'll squash all the commits into one and merge the changes 👌🏻

vjgarciag96:
Well, thinking about the squash... @Flame239 

I noticed from some previous PRs that if I do the squash and rebase from Github, it will add me as author in the commit and the commit will be authored by me and you (e.g. https://github.com/kotlin-telegram-bot/kotlin-telegram-bot/commit/978dce8914bd937e71d1c4e72c63703936823a70)

I'd like you to appear as the single author (because you are the author 😂), and the way to achieve that would be you doing the squash locally and pushing it to the PR. And then I'd have to just press the button and you'll appear as the single author.

Let me know if you can do that (there's no problem if you can't, we can always use the other option with mentioned downside).

Flame239:
I don't mind if you appear as author 🙂 
Let's merge it and go shopping for Christmas presents! 🎄

vjgarciag96:
Cool! It's now done, I finally achieved to make you appear as the author 💪🏻  (https://github.com/kotlin-telegram-bot/kotlin-telegram-bot/commit/0f4b3ee6754b8eca9c5ae5bf1e2e611b00724208).

Let's go for those Christmas presents! 🎁 🎅

Issue #141, opened 2021-01-14
Message handlers priorities and short circuiting


v1ctor:
Hi all,
I'm currently working on a bot which should support multiple session-based commands. I have my message handlers organized in this way:
```kotlin
dispatch {
                command("start") {
                    sendWelcome(message.chat.id)
                }
                command("find_mentors") {
                    val session = FindMentorsSession(mentorManager)
                    startSession(session, message.chat.id)
                }
                command("subscribe") {
                    val session = SubscribeSession()
                    startSession(session, message.chat.id)
                }
                text {
                    if (!text.startsWith("/")) {
                        val chatId = message.chat.id
                        handleUpdate(chatId, update)
                    }
                }
                callbackQuery {
                    val chatId = callbackQuery.message?.chat?.id ?: return@callbackQuery
                    handleUpdate(chatId, update)
                }
            }
```
The idea is that command handlers will create a new session while catch-all text and callback query handlers will forward update to the proper session.

The problem is that I have to filter out all the messages which start with / from text handler otherwise the message like `/subscribe` will be processed twice, once as command and again as a text message.
```kotlin
...
                    if (!text.startsWith("/")) {
...
```

My idea is to replace message handlers in the dispatcher from HashMap to the list to support a constant order of the handlers in the dispatcher, and also allow handlers to "consume" updates. It means that if one of the handlers consumed an update none of the handlers after that will get this update at all.

I'm quite happy to do it myself, I just want to get an early feedback on this idea before doing it.

Thanks in advance!

v1ctor:
@seik  please reassign it back to me if you're happy for me to do it.

vjgarciag96:
Hi @v1ctor! 

To filter out command messages you can also use the `message` handler with a `Text` filter instead of the `text` one:

```kotlin
dispatch {
    message(Filter.Text) {
        handleUpdate(message.chat.id, update)
    }
}
```

The idea sounds really interesting to me, happy for you to work on it! You can assign me as reviewer once the PR is ready for review 🙂

PR #142, opened 2021-01-19
Make update consumable

v1ctor:
Issue: #141 

### Summary
1. Update now inherits from ConsumableObject, so it can be marked as consumed by the handler.
2. Command handlers now use LinkedHashMap instead of HashMap to preserve and insertion order of handlers.

### Possible next steps
* Use numeric priorities for the handlers to sort them and don't rely on insertion order.

### As per contribution guidelines
```sh 
▶ ./gradlew ktlintFormat
BUILD SUCCESSFUL in 13s
6 actionable tasks: 6 executed

▶ ./gradlew check
BUILD SUCCESSFUL in 27s
16 actionable tasks: 15 executed, 1 up-to-date
```

### DIFF
--- a/docs/gettingUpdates.md
+++ b/docs/gettingUpdates.md
@@ -29,6 +29,23 @@ bot {
 
 For a more detailed example you can check the `dispatcher` and `echo` samples you can find in the `samples` folder of this project.
 
+### Update consumption
+Usually the `dispatcher` calls all the handlers which satisfy update parameters. If update was handled you can implicitly call `update.consume()` to mark the update as consumed so other handlers won't get it:  
+```kotlin
+bot { 
+    dispatcher {
+        text {  
+            if (text == 'Hello World!') {
+                update.consume()
+            }  
+        }
+        text {
+            // This handler will not get an update.
+        }
+    }
+}
+``` 
+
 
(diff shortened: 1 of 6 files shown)


v1ctor:
@vjgarciag96 can you please review it? Thanks!

vjgarciag96 (approved):
I wonder if it could make sense to implement this at a handler level to keep the `Update` objects as pure data objects without behaviour and immutable.

However, that would require a bit more thinking and lot of changes, so I'm happy to go with this approach for the moment, which is a simple implementation with a simple API that works properly.

Thank you so much for the contribution! 🙂 @v1ctor

I'll release a new version of the library with these changes

v1ctor:
> I wonder if it could make sense to implement this at a handler level to keep the `Update` objects as pure data objects without behaviour and immutable.
> 
> However, that would require a bit more thinking and lot of changes, so I'm happy to go with this approach for the moment, which is a simple implementation with a simple API that works properly.
> 
> Thank you so much for the contribution! 🙂 @v1ctor
> 
> I'll release a new version of the library with these changes

I've also considered other solutions but I didn't want to have breaking changes in the Handler's code. We can try to do that in the next major release. What do you think?

Issue #164, opened 2021-01-30
Add copyMessage method


vjgarciag96:
A new `copyMessage` method was added in Telegram Bot Api 5.0.

>Added the method copyMessage, which sends a copy of any message.

https://core.telegram.org/bots/api#copymessage

We should add this new method.

Flame239:
Let me take this one

Flame239:
@vjgarciag96  [please check this pr ](https://github.com/kotlin-telegram-bot/kotlin-telegram-bot/pull/170/files)

vjgarciag96:
Closed by #170

Issue #128, opened 2020-11-29
Add overloads for String type of chat_id


naftalmm:
In order to interact with public channels via Bot API, `chat_id` should be a `String` (in the format `@channelusername`). Currently, only few methods have such overload variant (with `channelUsername` argument name for `chat_id`, which is a bit confusing on my taste, but still better than nothing):
- [x] sendMessage
- [ ] forwardMessage (need overloads for `chat_id` & `from_chat_id` parameters, so it should be 4 methods summary)
- [ ] copyMessage (need overloads for `chat_id` & `from_chat_id` parameters, so it should be 4 methods summary)
- [ ] sendPhoto
- [ ] sendAudio
- [ ] sendDocument
- [ ] sendVideo
- [ ] sendAnimation
- [ ] sendVoice
- [ ] sendVideoNote
- [ ] sendMediaGroup
- [ ] sendLocation
- [ ] editMessageLiveLocation
- [ ] stopMessageLiveLocation
- [ ] sendVenue
- [ ] sendContact
- [x] sendPoll
- [x] sendDice
- [ ] sendChatAction
- [ ] kickChatMember
- [ ] unbanChatMember
- [ ] restrictChatMember
- [ ] promoteChatMember
- [x] setChatAdministratorCustomTitle
- [ ] setChatPermissions
- [ ] exportChatInviteLink
- [ ] setChatPhoto
- [ ] deleteChatPhoto
- [ ] setChatTitle
- [ ] setChatDescription
- [ ] pinChatMessage
- [ ] unpinChatMessage
- [ ] unpinAllChatMessages
- [ ] leaveChat
- [ ] getChat
- [ ] getChatAdministrators
- [ ] getChatMembersCount
- [ ] getChatMember
- [ ] setChatStickerSet
- [ ] deleteChatStickerSet
- [ ] editMessageText
- [ ] editMessageCaption
- [ ] editMessageMedia
- [ ] editMessageReplyMarkup
- [ ] stopPoll
- [x] deleteMessage
- [ ] sendSticker

Bunch of copy-paste should do the thing. Also there are 2 implementation options:
1. Preserve old or choose new name for `chat_id` parameter with `String` type. It should be consistent across all these methods. I believe it should be `chatId`, but it'll be a breaking change. Why `channelUsername` name was chosen initially on the first place?
2. Auto-add `@`-symbol in the beginning of `chat_id` string if it's not starting with it
```kotlin
[...]

PR #170, opened 2021-01-31
new api method: copy message

Flame239:
Implementation of `copyMessage` method introduced in Bot API 5.0 [(issue)](https://github.com/kotlin-telegram-bot/kotlin-telegram-bot/issues/164)
Also in this PR I tried a new way of working with `chatId` parameter (as discussed in https://github.com/kotlin-telegram-bot/kotlin-telegram-bot/issues/128)


telegram/src/main/kotlin/com/github/kotlintelegrambot/entities/ChatId.kt
```diff
@@ -0,0 +1,16 @@
+package com.github.kotlintelegrambot.entities
+
+/**
+ * Unique identifier for the target chat or username of the target channel (in the format @channelusername)
+ */
+sealed class ChatId {
+    data class Id(val id: Long) : ChatId()
+    class Username(username: String) : ChatId() {
```
vjgarciag96, line 8: Should we call this `ChannelUsername` instead?

telegram/src/test/kotlin/com/github/kotlintelegrambot/network/apiclient/ApiClientIT.kt
```diff
@@ -13,6 +14,8 @@ abstract class ApiClientIT {
 
     protected lateinit var sut: ApiClient
 
+    protected val gson = Gson()
```
vjgarciag96, line 17: I'm not really sure about this, I'd advocate for keeping this base test case as simple as possible and not start adding anything that is used by more than one test case here. 

A `Gson` instance is currently used in a couple of api client test cases only and there's not a great advantage from having it here.

telegram/src/test/kotlin/com/github/kotlintelegrambot/network/retrofit/converters/ChatIdConverterFactoryTest.kt
```diff
@@ -0,0 +1,52 @@
...
+        assertEquals(username, converter?.convert(Username(username)))
+    }
+
+    @Test
+    fun `returns a converter that transforms ChatId Username class to string and adds '@'`() {
+        val converter = sut.stringConverter(ChatId::class.java, emptyArray(), retrofitMock)
+
+        val username = "battlecruiser-operational"
+        assertEquals("@$username", converter?.convert(Username(username)))
+    }
```
vjgarciag96, line 51: This looks more like a test for `ChatId` than for the `Converter`? If we weren't using retrofit library and thus not using this converters we would still want to test that the @ is added to the channel username

telegram/src/main/kotlin/com/github/kotlintelegrambot/network/retrofit/converters/ChatIdConverterFactory.kt
```diff
@@ -0,0 +1,22 @@
...
+
+import com.github.kotlintelegrambot.entities.ChatId
+import com.github.kotlintelegrambot.entities.ChatId.Id
+import com.github.kotlintelegrambot.entities.ChatId.Username
+import java.lang.reflect.Type
+import retrofit2.Converter
+import retrofit2.Retrofit
+
+class ChatIdConverterFactory : Converter.Factory() {
+    override fun stringConverter(type: Type, annotations: Array<Annotation>, retrofit: Retrofit): Converter<ChatId, String>? {
```
vjgarciag96, line 11: Have you checked that the API operation works with both id types? I'm not 100% sure the Telegram Bot API will accept the numeric id as a string

vjgarciag96 (changes requested):

I left some comments, let me know if I can help with anything

telegram/src/test/kotlin/com/github/kotlintelegrambot/network/apiclient/CopyMessageIT.kt
```diff
@@ -0,0 +1,83 @@
...
+
+    private fun givenAnyCopyMessageResponse() {
+        val sendPollResponse = """
+            {"ok":true,"result":{"message_id":$ANY_RESULT_MESSAGE_ID}}
+        """.trimIndent()
+        val mockedResponse = MockResponse()
+            .setResponseCode(200)
+            .setBody(sendPollResponse)
+        mockWebServer.enqueue(mockedResponse)
+    }
```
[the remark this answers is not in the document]
Flame239: oh that copy paste 🤦

telegram/src/main/kotlin/com/github/kotlintelegrambot/network/retrofit/converters/ChatIdConverterFactory.kt
Flame239: Yes, I checked this. 
We use form urlencoded body and everything should be converted to string here (hence we override `stringConverter` method only).
If we decide to send json body where string and int representation are different, then we will need to use another deserialization approach.

Flame239:
@vjgarciag96 fixed 👌

telegram/src/main/kotlin/com/github/kotlintelegrambot/network/retrofit/converters/ChatIdConverterFactory.kt
vjgarciag96: Oh, got it. I guess we'll also need a gson serializer/deserializer for it at some point then.

vjgarciag96 (approved):
LGTM! Thanks for the contribution 🙂

PR #172, opened 2021-02-05
Chat id parameter refactoring

Flame239:
As discussed in https://github.com/kotlin-telegram-bot/kotlin-telegram-bot/issues/128, here I refactored all `chat_id` parameters to support numeric ids as well as channel usernames.

### DIFF
--- a/samples/polls/src/main/kotlin/com/github/kotlintelegrambot/polls/Main.kt
+++ b/samples/polls/src/main/kotlin/com/github/kotlintelegrambot/polls/Main.kt
@@ -4,6 +4,8 @@ import com.github.kotlintelegrambot.bot
 import com.github.kotlintelegrambot.dispatch
 import com.github.kotlintelegrambot.dispatcher.command
 import com.github.kotlintelegrambot.dispatcher.pollAnswer
+import com.github.kotlintelegrambot.entities.ChatId
+import com.github.kotlintelegrambot.entities.ChatId.Companion
 import com.github.kotlintelegrambot.entities.polls.PollType.QUIZ
 
 fun main() {
@@ -15,7 +17,7 @@ fun main() {
             }
             command("regularPoll") {
                 bot.sendPoll(
-                    chatId = message.chat.id,
+                    chatId = ChatId.fromId(message.chat.id),
                     question = "Pizza with or without pineapple?",
                     options = listOf("With :(", "Without :)"),
                     isAnonymous = false
@@ -24,7 +26,7 @@ fun main() {
 
             command("quizPoll") {
                 bot.sendPoll(
-                    chatId = message.chat.id,
+                    chatId = Companion.fromId(message.chat.id),
                     type = QUIZ,
                     question = "Java or Kotlin?",
                     options = listOf("Java", "Kotlin"),
@@ -35,7 +37,7 @@ fun main() {
 
             command("closedPoll") {
                 bot.sendPoll(
-                    chatId = message.chat.id,
+                    chatId = Companion.fromId(message.chat.id),
                     type = QUIZ,
                     question = "Foo or Bar?",
                     options = listOf("Foo", "Bar", "FooBar"),
--- a/samples/polls/src/main/kotlin/com/github/kotlintelegrambot/polls/Main.kt
+++ b/samples/polls/src/main/kotlin/com/github/kotlintelegrambot/polls/Main.kt
@@ -5,7 +5,6 @@ import com.github.kotlintelegrambot.dispatch
 import com.github.kotlintelegrambot.dispatcher.command
 import com.github.kotlintelegrambot.dispatcher.pollAnswer
 import com.github.kotlintelegrambot.entities.ChatId
-import com.github.kotlintelegrambot.entities.ChatId.Companion
 import com.github.kotlintelegrambot.entities.polls.PollType.QUIZ
 
 fun main() {
@@ -26,7 +25,7 @@ fun main() {
 
             command("quizPoll") {
                 bot.sendPoll(
-                    chatId = Companion.fromId(message.chat.id),
+                    chatId = ChatId.fromId(message.chat.id),
                     type = QUIZ,
                     question = "Java or Kotlin?",
                     options = listOf("Java", "Kotlin"),
@@ -37,7 +36,7 @@ fun main() {
 
             command("closedPoll") {
                 bot.sendPoll(
-                    chatId = Companion.fromId(message.chat.id),
+                    chatId = ChatId.fromId(message.chat.id),
                     type = QUIZ,
                     question = "Foo or Bar?",
                     options = listOf("Foo", "Bar", "FooBar"),
--- a/telegram/src/main/kotlin/com/github/kotlintelegrambot/Bot.kt
+++ b/telegram/src/main/kotlin/com/github/kotlintelegrambot/Bot.kt
@@ -257,7 +257,7 @@ class Bot private constructor(
     ).call()
 
     fun sendPhoto(
-        chatId: Long,
+        chatId: ChatId,
         photo: SystemFile,
         caption: String? = null,
         parseMode: ParseMode? = null,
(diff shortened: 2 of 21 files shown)


Flame239:
@vjgarciag96 please take a look

vjgarciag96:
Thanks for the PR! I'll take a look to it probably tomorrow. 

Next time you can just assign me as a reviewer in the PR directly if you prefer 🙂

Flame239:
Unfortunately I can't assign you a reviewer, don't have permissions

vjgarciag96:
Oh, true. I added you as a collaborator to the repo, if you accept the invite I think you'll be able to request my review on next PRs 🙂

samples/polls/src/main/kotlin/com/github/kotlintelegrambot/polls/Main.kt
```diff
@@ -24,7 +26,7 @@ fun main() {
 
             command("quizPoll") {
                 bot.sendPoll(
-                    chatId = message.chat.id,
+                    chatId = Companion.fromId(message.chat.id),
```
vjgarciag96, line 29:
```suggestion
                    chatId = ChatId.fromId(message.chat.id),
```

samples/polls/src/main/kotlin/com/github/kotlintelegrambot/polls/Main.kt
```diff
@@ -35,7 +37,7 @@ fun main() {
 
             command("closedPoll") {
                 bot.sendPoll(
-                    chatId = message.chat.id,
+                    chatId = Companion.fromId(message.chat.id),
```
vjgarciag96, line 40:
```suggestion
                    chatId = ChatId.fromId(message.chat.id),
```

samples/polls/src/main/kotlin/com/github/kotlintelegrambot/polls/Main.kt
```diff
@@ -4,6 +4,8 @@ import com.github.kotlintelegrambot.bot
 import com.github.kotlintelegrambot.dispatch
 import com.github.kotlintelegrambot.dispatcher.command
 import com.github.kotlintelegrambot.dispatcher.pollAnswer
+import com.github.kotlintelegrambot.entities.ChatId
+import com.github.kotlintelegrambot.entities.ChatId.Companion
```
vjgarciag96, line 8:
```suggestion

```

telegram/src/main/kotlin/com/github/kotlintelegrambot/Bot.kt
```diff
@@ -203,20 +203,8 @@ class Bot private constructor(
...
-     * @param chatId unique identifier for the target chat
-     * @param text text of the message to be sent, 1-4096 characters after entities parsing
-     * @param parseMode mode for parsing entities in the message text
-     * @param disableWebPagePreview disables link previews for links in this message
-     * @param disableNotification sends the message silently - users will receive a notification with no sound
-     * @param replyToMessageId if the message is a reply, ID of the original message
-     * @param replyMarkup additional options - inline keyboard, custom reply keyboard, instructions to remove reply
-     * keyboard or to force a reply from the user
-     * @return the sent [Message] on success
-     */
```
vjgarciag96, line 217: I think we should keep the documentation in this method, and just update the description of the `chatId` parameter to reflect that now it can be either a chat id or a channel username:

```kotlin
* @param chatId Unique identifier for the target chat or username of the target channel (in
* the format @channelusername)
```

telegram/src/main/kotlin/com/github/kotlintelegrambot/entities/ChatId.kt
```diff
@@ -4,9 +4,13 @@ package com.github.kotlintelegrambot.entities
...
 sealed class ChatId {
-    data class Id(val id: Long) : ChatId()
+    data class Id(val id: Long) : ChatId() {
+        override fun toString() = id.toString()
+    }
+
     class ChannelUsername(username: String) : ChatId() {
         val username: String = if (username.startsWith("@")) username else "@$username"
+        override fun toString() = username
     }
```
vjgarciag96, line 14: Hmm, I'm not a sure about relying on the `toString` method to transform a `ChatId` for multipart body requests. 

What do you think about having an extension function like the next encapsulating that transformation instead:

```kotlin
fun ChatId.toMultipartBodyPart(partName: String): MultipartBody.Part = when(this) {
    is Id -> id.toMultipartBodyPart(partName)
    is ChannelUsername -> username.toMultipartBodyPart(partName)
}
```

We can add it in the `MultipartExtensions.kt` file so it's grouped with other similar functions.

vjgarciag96 (commented):

It looks good to me overall, I just left some little comments. Let me know if I can help with anything 🙂

samples/polls/src/main/kotlin/com/github/kotlintelegrambot/polls/Main.kt
Flame239: 🤔  where does this come from

telegram/src/main/kotlin/com/github/kotlintelegrambot/Bot.kt
Flame239: 👍  updated param description in other methods also

telegram/src/main/kotlin/com/github/kotlintelegrambot/entities/ChatId.kt
Flame239: Reason why I don't like such extension method usage: 
converting to MultipartBody is totally none of the business of ChatId class, so it shouldn't be a part of this class. Also creating extension methods decreases code readability, because instead of looking in 1 file with class definition, developer should looks in multiple files (which he should find first).

`toString()` here is doing exactly what is was made for: returns a string representation of the object. At the end of the day, method with converts object to string RequestBody expects string representation of the object as input, which is `toString()`

vjgarciag96: The extension function I was proposing in the comment is not a thing of `ChatId` class. I don't want it to be part of the class, that's why I'm proposing the implementation with an extension function and not with a member function. Using extension functions to map a class to a different representation of that class is a common pattern in Kotlin. I don't see how that decreases code readability. If we had to implement the same without extension functions we'll probably create a class with a single function that will receive the `ChatId` as parameter and will return a `String`, and that will be harder to read and use in an IDE than the extension function.

`toString()` says that the returned string should be "a concise but informative representation that is easy for a person to read", and that's true with your changes. Why I'm not in favour of using `toString` for this then?:

* If this value returned by `toString()` is used when we need to map the chat id value for multipart parts and clients widely use it, we'll be stuck with it, and if we change it we'll break all these clients. In this case we're the only clients supposed to do this, so it's not such a big deal, but we'll be limiting ourselves to be stuck with this `toString` implementation.
* If we wanted a different human readable representation returned by `toString` for this class for debugging purposes, we couldn't have it because we would break the multipart code.
* Currently, we don't override the `toString()` for other Telegram Bot Api classes and rely on the `toString` generated by data classes (actually there are some classes that override it with the same idea you're proposing here too, but we want to move them away from that). So it won't be consistent with the `toString` of other classes representing Telegram Bot API objects, which is not a big deal either, but could be annoying to library clients.
* If in the future we need to add a new subclass for the `ChatId` sealed class and we forget to override the `toString`, our code will silently fail. If we do that and we're using the extension function approach, the code won't compile until we consider the new subclass in our mapping code.

What do you think about this?

Flame239: Alright, your arguments are quiet reasonable, although I found a better way of doing it 🙃 
`ChatIdConverterFactory` now implements `requestBodyConverter()` which converts class to `RequestBody` under `@Part` annotation in `@Multipart` requests.

vjgarciag96 (approved):
Nice one! LGTM

Thank you so much for the contribution! 🙂

Issue #146, opened 2021-01-30
Add logOut method


vjgarciag96:
A new `logOut` method was added in Telegram Bot Api 5.0:

> Use this method to log out from the cloud Bot API server before launching the bot locally. You must log out the bot before running it locally, otherwise there is no guarantee that the bot will receive updates. After a successful call, you can immediately log in on a local server, but will not be able to log in back to the cloud Bot API server for 10 minutes. Returns True on success. Requires no parameters.

 https://core.telegram.org/bots/api#logout

We should add support for this method.

vjgarciag96:
Closed by #185

PR #185, opened 2021-02-27
Add logOut method and tests

MyMindWorld:
issue #146
Also i allowed myself to fix some of non-critical issues with documentation and deprecated `classifier` 

Assertj in my opinion seems more convinient for openSource, due to clear assertion visibility.

As for tests, if im not only one having them failing, i have testFixes branch

### DIFF
--- a/telegram/build.gradle
+++ b/telegram/build.gradle
@@ -34,12 +34,13 @@ dependencies {
     testImplementation "org.junit.jupiter:junit-jupiter:$junit_version"
     testImplementation "org.junit.jupiter:junit-jupiter-api:$junit_version"
     testImplementation "org.junit.jupiter:junit-jupiter-params:$junit_version"
+    testImplementation "org.assertj:assertj-core:3.11.1"
     testImplementation "io.mockk:mockk:1.10.2"
     testImplementation "com.squareup.okhttp3:mockwebserver:$mock_web_server_version"
 }
 
 task sourcesJar(type: Jar, dependsOn: classes) {
-    classifier = 'sources'
+    archiveClassifier.set("sources")
     from sourceSets.main.allSource
 }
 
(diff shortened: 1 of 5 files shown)


telegram/build.gradle
```diff
@@ -34,12 +34,13 @@ dependencies {
...
     testImplementation "org.junit.jupiter:junit-jupiter-api:$junit_version"
     testImplementation "org.junit.jupiter:junit-jupiter-params:$junit_version"
+    testImplementation "org.assertj:assertj-core:3.11.1"
     testImplementation "io.mockk:mockk:1.10.2"
     testImplementation "com.squareup.okhttp3:mockwebserver:$mock_web_server_version"
 }
 
 task sourcesJar(type: Jar, dependsOn: classes) {
-    classifier = 'sources'
+    archiveClassifier.set("sources")
```
MyMindWorld, line 43: https://discuss.gradle.org/t/publishing-jars-with-classifiers-after-v5-0/30144/5

telegram/src/main/kotlin/com/github/kotlintelegrambot/Bot.kt
```diff
@@ -1255,7 +1266,7 @@ class Bot private constructor(
 
     /**
      * Once the user has confirmed their payment and shipping details, the Bot API sends the final
-     * confirmation in the form of an [Update] with the field [preCheckoutQuery]. Use this method to
+     * confirmation in the form of an [Update] with the field [preCheckoutQueryId]. Use this method to
```
MyMindWorld, line 1269: With `id` this links become parsable for IDEA to point on field

vjgarciag96 (approved):
Thank you so much for the contribution @MyMindWorld!

Let's give a try to the `assertj` library and see how it goes. I guess there are similar libraries built for Kotlin rather than Java, but let's see the value this one brings and then we could consider the next step 👍🏻

Issue #167, opened 2021-01-30
Add allow_sending_without_reply to several methods


vjgarciag96:
A new `allow_sending_without_replay` parameter was added to the methods sendMessage, sendPhoto, sendVideo, sendAnimation, sendAudio, sendDocument, sendSticker, sendVideoNote, sendVoice, sendLocation, sendVenue, sendContact, sendPoll, sendDice, sendInvoice, sendGame and sendMediaGroup in Telegram Bot Api 5.0.

>Added the field allow_sending_without_reply to the methods sendMessage, sendPhoto, sendVideo, sendAnimation, sendAudio, sendDocument, sendSticker, sendVideoNote, sendVoice, sendLocation, sendVenue, sendContact, sendPoll, sendDice, sendInvoice, sendGame, sendMediaGroup to allow sending messages not a as reply if the replied-to message has already been deleted.

We should add this new parameter to all those methods.

vjgarciag96:
Mostly done in #186.

There are a couple of remaining tasks related to it, they're documented in #187 and #188.

PR #186, opened 2021-02-28
add `allow_sending_without_reply` param to methods

MyMindWorld:
Issue #167

Besides i removed some unused params i found along the way and refactored hardcoded strings to ApiConstants

### DIFF
--- a/telegram/src/test/kotlin/com/github/kotlintelegrambot/network/apiclient/SendMediaGroupIT.kt
+++ b/telegram/src/test/kotlin/com/github/kotlintelegrambot/network/apiclient/SendMediaGroupIT.kt
@@ -41,7 +41,7 @@ class SendMediaGroupIT : ApiClientIT() {
         val expectedRequestBody = String.format(
             getFileAsStringFromResources<SendMediaGroupIT>("sendMediaGroupRequestBody1.txt"),
             multipartBoundary
-        )
+        ).trimIndent()
         assertEquals(expectedRequestBody, requestBody)
     }
 
@@ -61,7 +61,7 @@ class SendMediaGroupIT : ApiClientIT() {
         val expectedRequestBody = String.format(
             getFileAsStringFromResources<SendMediaGroupIT>("sendMediaGroupRequestBody2.txt"),
             multipartBoundary
-        )
+        ).trimIndent()
         assertEquals(expectedRequestBody, requestBody)
     }
 
@@ -81,7 +81,7 @@ class SendMediaGroupIT : ApiClientIT() {
         val expectedRequestBody = String.format(
             getFileAsStringFromResources<SendMediaGroupIT>("sendMediaGroupRequestBody3.txt"),
             multipartBoundary
-        )
+        ).trimIndent()
         assertEquals(expectedRequestBody, requestBody)
     }
 
@@ -104,7 +104,7 @@ class SendMediaGroupIT : ApiClientIT() {
         val expectedRequestBody = String.format(
             getFileAsStringFromResources<SendMediaGroupIT>("sendMediaGroupRequestBody4.txt"),
             multipartBoundary
-        )
+        ).trimIndent()
         assertEquals(expectedRequestBody, requestBody)
     }
 
@@ -124,7 +124,7 @@ class SendMediaGroupIT : ApiClientIT() {
         val expectedRequestBody = String.format(
             getFileAsStringFromResources<SendMediaGroupIT>("sendMediaGroupRequestBody5.txt"),
             multipartBoundary
-        )
+        ).trimIndent()
         assertEquals(expectedRequestBody, requestBody)
     }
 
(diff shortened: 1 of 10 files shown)


telegram/src/test/kotlin/com/github/kotlintelegrambot/network/apiclient/SendMediaGroupIT.kt
```diff
@@ -41,7 +41,7 @@ class SendMediaGroupIT : ApiClientIT() {
         val expectedRequestBody = String.format(
             getFileAsStringFromResources<SendMediaGroupIT>("sendMediaGroupRequestBody1.txt"),
             multipartBoundary
-        )
+        ).trimIndent()
```
vjgarciag96, line 44: Why is this change needed?

telegram/src/test/kotlin/com/github/kotlintelegrambot/network/apiclient/SetWebhookIT.kt
```diff
@@ -96,7 +96,7 @@ class SetWebhookIT : ApiClientIT() {
         val expectedRequestBody = String.format(
             getFileAsStringFromResources<SetWebhookIT>("setWebhookMultipartWithCertificateRequestBody.txt"),
             multipartBoundary
-        )
+        ).trimIndent()
```
vjgarciag96, line 99: Same as before - why is this change needed?

telegram/src/test/kotlin/com/github/kotlintelegrambot/types/TelegramBotResultTest.kt
```diff
@@ -61,7 +61,7 @@ class TelegramBotResultTest {
     }
```
vjgarciag96, line 61: I would keep the previous test names. One test name or other is mostly a matter of taste, and is a change unrelated to the PR, so I would advocate to keep it as was. Otherwise, we could end up with back and forth changes in tests names that lead nowhere.

vjgarciag96 (commented):
Thank you so much for the contribution @MyMindWorld, good job!

I added a couple of minor comments in the PR.

I can see that there are two additional API methods that also have this new parameter, `sendGame` and `sendMediaGroup`. We don't have the `sendGame` method in the library yet, so I have created a new issue to keep that in the radar (https://github.com/kotlin-telegram-bot/kotlin-telegram-bot/issues/187). And the `sendMediaGroup` is a bit special compared to the other methods, so I've also created an issue to add the parameter there as a separated task (https://github.com/kotlin-telegram-bot/kotlin-telegram-bot/issues/188).

Regarding the PR, I like the cleanup changes about using the API constants. It would be nice if we could keep the cleanup commits separated from the others for clarity, to keep a better git history and to make reviewing easier. I mean, it's fine and we can go ahead with them as they're here, but I'm just commenting this in case it can be helpful for future PRs.

Thanks again for the contribution! 🙂

telegram/src/test/kotlin/com/github/kotlintelegrambot/network/apiclient/SendMediaGroupIT.kt
MyMindWorld: On windows machine we receive content with indent in the beginning, In real-life application they won't matter. 
image
If there is more to dig, please point where to)

telegram/src/test/kotlin/com/github/kotlintelegrambot/types/TelegramBotResultTest.kt
MyMindWorld: What do you think than of separate pr with naming convention?
I belive that tests naming should follow one 
e.g. `Should do something when given` or `When given should do something`
This way we can detect some issues with architecrure or split tests with similar setups into nested classes
e.g.
```kotlin
@Nested
inner class `Should send MediaGroup with media group composed by a video(fileUrl), a photo(fileId)` {
    // shared fixture with response and media group
    @Test
    fun `when all mandatory arhuments is passed`() {
        // test
    }
    @Test
    fun `when all arguments is passed`() {
        // test
    }
```

vjgarciag96: I don't have a strong opinion about tests naming. I used to follow the "should do something when given" naming but it sometimes ends up being too verbose and makes test names and tests harder to understand. Now I try to keep it simple in the name and just follow a `given-when-then`|`arrange-act-assert` structure in the test itself. I guess it's also a matter of taste and something to be agreed with all the people working in a project to be used in a consistent way (in this case the project is quite small and the people working on it is not too many, so I guess it's not that critical).

I like the nested tests though, we're not taking advantage of them and we should, good call 👍🏻

vjgarciag96 (approved):
Let's merge this!

telegram/src/test/kotlin/com/github/kotlintelegrambot/network/apiclient/SendMediaGroupIT.kt
vjgarciag96: Oh gotcha, good to know!

telegram/src/test/kotlin/com/github/kotlintelegrambot/types/TelegramBotResultTest.kt
MyMindWorld: Maybe than we should create issue with poll about it? Also it would help with writing contributing guide for project, it could decrease number of this **taste** cases
Thanks for review!)

Issue #187, opened 2021-02-28
Add sendGame method


vjgarciag96:
This method hasn't been added to the library yet.

https://core.telegram.org/bots/api#sendgame

vjgarciag96:
Closed by #192

PR #192, opened 2021-03-06
Adds sendGame method - Issue #187

LeGorge:
Adds sendGame support - Issue #187 

- creates junit tests for the new method
 - updates documentation
 - adds the entity CallbackGame which is used by the telegram api as a placeholder
and as of now holds no information
 - adds support for the inlineKeyboardButton that is supposed to be of CallbackGame type
 - adds the field replyMarkup to the Message example factory(Mothers.kt) in the tests
 - sets the default values of errorCode and errorDescription in Response Serialization data class
to null so it is not required to pass them when instatiating

Also ktlint changed a few imports in Mother.kt.

Please let me know if you have any questions, suggestions or changes. :)

### DIFF
--- a/telegram/src/main/kotlin/com/github/kotlintelegrambot/entities/InlineKeyboardMarkup.kt
+++ b/telegram/src/main/kotlin/com/github/kotlintelegrambot/entities/InlineKeyboardMarkup.kt
@@ -14,12 +14,22 @@ data class InlineKeyboardMarkup internal constructor(
 ) : ReplyMarkup {
 
     init {
-        // Buttons of type [Pay] must always be the first button in the first row.
+        validatePriorityButtonsForType<InlineKeyboardButton.Pay>()
+        validatePriorityButtonsForType<InlineKeyboardButton.CallbackGameButtonType>()
+    }
+
+    // Priority buttons must always be the first button in the first row.
+    private inline fun <reified T> validatePriorityButtonsForType() {
         val flattenedButtons = inlineKeyboard.flatten()
-        val payButtons = flattenedButtons.filterIsInstance<InlineKeyboardButton.Pay>()
-        require(payButtons.size <= 1) { "Can't have more than one pay button per inline keyboard" }
-        require(payButtons.size != 1 || flattenedButtons.firstOrNull() is InlineKeyboardButton.Pay) {
-            "Pay buttons must always be the first button in the first row"
+        val filteredButtons = flattenedButtons.filterIsInstance<T>()
+        val typeName = T::class.simpleName
+
+        require(filteredButtons.size <= 1) {
+            "Can't have more than one button of type $typeName per inline keyboard"
+        }
+
+        require(filteredButtons.size != 1 || flattenedButtons.firstOrNull() is T) {
+            "Buttons of type $typeName must always be the first button in the first row"
         }
     }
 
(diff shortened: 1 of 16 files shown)


LeGorge:
The task telegram:apiCheck broke the pipeline.
I confess I have no idea what this was about, but calling ./gradlew telegram:apiDump as suggested in the error seems to have fixed the problem.
Perhaps this is something that needs to be documented somewhere.

telegram/src/main/kotlin/com/github/kotlintelegrambot/entities/keyboard/InlineKeyboardButton.kt
```diff
@@ -45,6 +46,15 @@ sealed class InlineKeyboardButton {
         @SerializedName("switch_inline_query_current_chat") val switchInlineQueryCurrentChat: String
     ) : InlineKeyboardButton()
 
+    /**
+     * Description of the game that will be launched when the user presses the button.
+     * NOTE: this type of button must always be the first button in the first row.
```
vjgarciag96, line 51: Pay buttons have the same constraint and to ensure library users don't use them incorrectly, there is some validation code in `InlineKeyboardMarkup` to ensure that pay buttons are the first button in the first row. Should we do the same for this type of button?

vjgarciag96 (approved):
Thank you so much for the contribution @LeGorge! LGTM, good job! Love the documentation 👍🏻

I only added a little comment in the PR.

Regarding the `apiCheck` and `apiDump` thing, it's just a new plugin we added that generates a file with the public binary api of the library and checks that it doesn't change if we add new changes.

Basically the idea of having it is to not publish things that we don't want to expose to library clients and also ensure we don't break the public API accidentally. So every time you intentionally add a change in the public binary API of the library you'll need to manually run `apiDump` to regenerate the file with the new change. It's integrated in the CI so we can catch any accidental change in the public API automatically in the PRs. I'm still not sure if it will add enough value to the library or if it will be more noise and overhead, but it's worth a try (and I think it'll be helpful to force to think about what things should be exposed for library clients). I'll add a little explanation about it in the documentation.

telegram/src/main/kotlin/com/github/kotlintelegrambot/entities/keyboard/InlineKeyboardButton.kt
LeGorge: @vjgarciag96 Ow I missed that, it's been fixed.
Also I took the liberty of refactoring the tests for this constraint with a Mothers factory for inlineKeyboardButtons.

telegram/src/main/kotlin/com/github/kotlintelegrambot/entities/InlineKeyboardMarkup.kt
```diff
@@ -14,12 +14,22 @@ data class InlineKeyboardMarkup internal constructor(
...
+
+    // Priority buttons must always be the first button in the first row.
+    private inline fun <reified T> validatePriorityButtonsForType() {
         val flattenedButtons = inlineKeyboard.flatten()
-        val payButtons = flattenedButtons.filterIsInstance<InlineKeyboardButton.Pay>()
-        require(payButtons.size <= 1) { "Can't have more than one pay button per inline keyboard" }
-        require(payButtons.size != 1 || flattenedButtons.firstOrNull() is InlineKeyboardButton.Pay) {
-            "Pay buttons must always be the first button in the first row"
+        val filteredButtons = flattenedButtons.filterIsInstance<T>()
+        val typeName = T::class.simpleName
```
vjgarciag96, line 25: Not really important: Should we access the class name only when it's needed and not beforehand? I guess most of the cases will be valid inline keyboards that don't need to load the button type name.

I mean, I don't think it's going to make a difference. I'm happy to merge this as is (I'll leave the comment for some minutes and I'll merge it if you don't see the comment in that time).

vjgarciag96 (approved):
Looks great! Thank you so much for the contribution again @LeGorge 👏🏻

vjgarciag96:
Merging!

PR #200, opened 2021-03-13
Add replyToMessage extension to all applicable HandlerEnvironments.

findusl:
In my own code I found a simple replyToMessage extension very handy in the HandlerEnvironments. However I have to create one for every HandlerEnvironment. I think to generalize the Environments with some Interfaces would solve that problem.

I'm not sure if the filenames and locations that I chose here are really the most suitable ones though. I don't know the project very well so far and I'm rather new to Kotlin too. I would love some feedback on that so I can improve.

This is the first time I really try to contribute on Github. If I make any mistakes in the procedure please tell me :)

### DIFF
--- a/telegram/api/telegram.api
+++ b/telegram/api/telegram.api
@@ -240,7 +240,7 @@ public final class com/github/kotlintelegrambot/dispatcher/handlers/ChannelHandl
 	public fun toString ()Ljava/lang/String;
 }
 
-public final class com/github/kotlintelegrambot/dispatcher/handlers/CommandHandlerEnvironment {
+public final class com/github/kotlintelegrambot/dispatcher/handlers/CommandHandlerEnvironment : com/github/kotlintelegrambot/dispatcher/handlers/IMessageHandlerEnvironment {
 	public final fun component1 ()Lcom/github/kotlintelegrambot/Bot;
 	public final fun component2 ()Lcom/github/kotlintelegrambot/entities/Update;
 	public final fun component3 ()Lcom/github/kotlintelegrambot/entities/Message;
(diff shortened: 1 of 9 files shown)


vjgarciag96 (commented):
@findusl thanks for the contribution! (and sorry for the late reply) I see your point about the extension, but I think it would be better to keep the core library API simpler and not add these specific extensions as part of it.

In one hand, the function is just a bit of syntactic sugar over `bot.sendMessage`, and if we start adding similar functions the API of the library would grow a lot (we could add a function like this for almost every API operation), making it more complex to use with not much benefit. Also, if extensions like this are useful for your use case, it's really easy to implement them in your app thanks to extension functions. On the other hand, if some of the library users find functions like this especially useful, it could be a good idea to provide an additional extension library (separated from the core library) with them. In this way, users that find them useful can easily add the new dependency and take advantage of them, and users that don't want them can use the library in the same way they've been always doing.

What do you think about this?

KPidS:
When I started using the library I immediately thought "Why isn't this a thing". You can add your own extension functions, but because the classes don't have a common interface – you are forced to repeat functions with the same logic. It's a bad design in the library that forces you to make bad design down the line in the extension functions

seik:
Hi @findusl — thanks for the contribution and again, apologies for the very long wait. @vjgarciag96's reasoning still applies: a single `replyToMessage` extension on every handler environment would grow the core API surface significantly for what's mostly syntactic sugar over `bot.sendMessage(...)`, and once you start there it's hard to draw the line for similar helpers.

The cleaner path for projects that want it is to add it as a local extension in your own codebase — and if having a shared interface across HandlerEnvironments makes that easier, that's worth a focused separate issue.

Closing this one. Really appreciate the original work!

Issue #208, opened 2021-06-08
Unify all Bot API related to sending files (use TelegramFile type everythere)


naftalmm:
**Is your feature request related to a problem? Please describe.**
`TelegramFile` type encapsulates the way of [sending files](https://core.telegram.org/bots/api#sending-files): via file_id/URL or actual file uploading.
Currently this class is used for creating other objects, passed as parameters into `editMessageMedia` and `sendMediaGroup` API methods.
All other methods related to sending files have overloaded versions for parameters of `String` / `java.io.File` types (where `String` is used to represent file_id or URL).
Sometimes it's not known at the start of request processing, what will be the response - a media group (`sendMediaGroup` will be called), or a single media (some other API method), so it would be handful to prepare media files in a common way, being sured that all API methods will accept it.

**Describe the solution you'd like**
Replace 2 overloaded variants for [sendPhoto](https://core.telegram.org/bots/api#sendphoto), [sendAudio](https://core.telegram.org/bots/api#sendaudio), [sendDocument](https://core.telegram.org/bots/api#senddocument), [sendVideo](https://core.telegram.org/bots/api#sendvideo), [sendAnimation](https://core.telegram.org/bots/api#sendanimation), [sendVoice](https://core.telegram.org/bots/api#sendvoice) methods with one, using parameter of `TelegramFile` type. This will make API clearer, more consistent and will allow to better emphasize ways of input file representation.
Sending video notes by a URL is currently unsupported, so [sendVideoNote](https://core.telegram.org/bots/api#sendvideonote) methods could be kept as is (or replaced with overloads for `TelegramFile.ByFileId` and `TelegramFile.ByFile`).

**Describe alternatives you've considered**
Bot API could be simplified even further if `InputMediaPhoto` `InputMediaAudio`, `InputMediaDocument`, `InputMediaVideo`, `InputMediaAnimation` types (they all have a field of `TelegramFile` type) were used instead of bunch of parameters (logically grouping them) in [sendPhoto](https://core.telegram.org/bots/api#sendphoto), [sendAudio](https://core.telegram.org/bots/api#sendaudio), [sendDocument](https://core.telegram.org/bots/api#senddocument), [sendVideo](https://core.telegram.org/bots/api#sendvideo), [sendAnimation](https://core.telegram.org/bots/api#sendanimation) methods respectfully.

**Implementation ideas**
[...]

PR #203, opened 2021-04-15
Adds disable_content_type_detection to sendDocument and InputMediaDoc…

LeGorge:
…ument

 - Updates sendDocument to use the TelegramBotResult
 - Adds tests for sendDocument and some resource files for them
 - Adds mimeType param to sendDocument(ByteArray) as this is necessary for Content-Type expliciting
 - Updates the public library api
 - Closes issue #153

PR #210, opened 2021-06-18
Unify sending files API

naftalmm:
This PR resolves issue #208


telegram/src/test/kotlin/com/github/kotlintelegrambot/testutils/IOUtils.kt
```diff
@@ -1,9 +1,10 @@
...
 import java.io.File
+import java.net.URLDecoder
 
 inline fun <reified T> getFileAsStringFromResources(resName: String): String =
-    T::class.java.getResource("/$resName").readText()
+    getFileFromResources<T>(resName).readText()
 
 inline fun <reified T> getFileFromResources(resName: String): File =
-    File(T::class.java.getResource("/$resName").file)
+    File(URLDecoder.decode(T::class.java.getResource("/$resName")!!.file, "UTF-8"))
```
vjgarciag96, line 10: Do we need the `URLDecoder.decode` part?

telegram/src/main/kotlin/com/github/kotlintelegrambot/network/serialization/adapter/TelegramFileAdapter.kt
```diff
@@ -14,5 +15,6 @@ internal class TelegramFileAdapter : JsonSerializer<TelegramFile> {
         is ByFileId -> context.serialize(src.fileId, String::class.java)
         is ByUrl -> context.serialize(src.url, String::class.java)
         is ByFile -> context.serialize("attach://${src.file.name}")
+        is ByByteArray -> context.serialize("attach://${src.filename!!}")
```
vjgarciag96, line 18: Won't this crash if the `ByByteArray` object doesn't have a filename? Should we enforce a filename/provide a fallback for it?

vjgarciag96 (commented):
@naftalmm thank you so much for the contribution, really nice job! (and sorry for the late review).

Apart from a couple of minor things, LGTM

vjgarciag96:
I reckon there will be some merge conflicts with https://github.com/kotlin-telegram-bot/kotlin-telegram-bot/pull/203. I can help with those if you want @naftalmm

telegram/src/main/kotlin/com/github/kotlintelegrambot/network/serialization/adapter/TelegramFileAdapter.kt
naftalmm: This case is covered in [`sendVoice with audio from ByteArray`](https://github.com/kotlin-telegram-bot/kotlin-telegram-bot/blob/46d2e8c603c1379fe9c76d6919a2ba39c0f7d32d/telegram/src/test/kotlin/com/github/kotlintelegrambot/network/apiclient/SendVoiceIT.kt#L89) test, which passes.

Actually, this code line is never called (as well as preceding; it's here only for `when` exhaustiveness), because “attach://<file_attach_name>” is never sent as a JSON field, it's used only with "multipart/form-data" (fallback for the filename is provided [during conversion to `MultipartBody.Part`](https://github.com/kotlin-telegram-bot/kotlin-telegram-bot/blob/46d2e8c603c1379fe9c76d6919a2ba39c0f7d32d/telegram/src/main/kotlin/com/github/kotlintelegrambot/network/multipart/MultipartExtensions.kt#L34)).

telegram/src/test/kotlin/com/github/kotlintelegrambot/testutils/IOUtils.kt
naftalmm: Yes, otherwise if the path to the project folder on developer machine contains spaces, tests calling this method will fail `java.io.FileNotFoundException`:

>E:\Git%20repositories\kotlin-telegram-bot\telegram\build\resources\test\short.ogg (Системе не удается найти указанный путь)

naftalmm: I just realized that Kotlin stdlib has a nicer way to do this:

```kt
T::class.java.getResource("/$resName")!!.file.decode()
```

telegram/src/main/kotlin/com/github/kotlintelegrambot/network/serialization/adapter/TelegramFileAdapter.kt
vjgarciag96: It's a bit tricky, but I think this code path could be reached from the `sendMediaGroup` operation https://github.com/kotlin-telegram-bot/kotlin-telegram-bot/blob/1773290206d217999c73a88bbeec824ba591b50f/telegram/src/main/kotlin/com/github/kotlintelegrambot/network/multipart/MultipartBodyFactory.kt#L39. If you check the code there we use `Gson` to deserialise "groupable" media which ends up deserialising `TelegramFile`s.

However, we don't support `ByteArray` files in that API yet (although a library client is allowed to pass `ByteArray` files to it), so that would need a bit of extra work. We can leave it as is for now and I'll work on the changes to support this properly for `sendMediaGroup` before the next release of the library is shipped if that sounds good to you.

vjgarciag96 (approved):

Issue #213, opened 2021-08-05
editMessageMedia does not function


zhtaihao:
**Expected behavior**
[editMessageMedia](https://github.com/kotlin-telegram-bot/kotlin-telegram-bot/blob/main/telegram/src/main/kotlin/com/github/kotlintelegrambot/Bot.kt#L1220) edits the media of a message. The user sees this change immediately.

**Actual behavior**
Nothing happens.

**To Reproduce**
Call the method linked above and provide a bot-unique file ID.

**Screenshots/Traceback**

**Possible solutions**
Perhaps this is related to the JSON serialisation of the TelegramFile.ByFileId object? [[reference]](https://github.com/kotlin-telegram-bot/kotlin-telegram-bot/blob/main/telegram/src/main/kotlin/com/github/kotlintelegrambot/network/ApiService.kt#L643)

**Additional context**
Using the same parameters in a direct POST request to the API endpoint (http://api.telegram.org/...), the message gets edited correctly. This validates message ID, chat ID, and file ID.

zhtaihao:
We haven't migrated to version 6.0.4 yet. This bug is observed in 6.0.2.

rastiehaiev:
I face the same problem. When I try to edit message media, Telegram returns 400 with error 
`{"ok":false,"error_code":400,"description":"Bad Request: can't parse input media JSON object"}`. I looked at the code and everything seems to be fine.

rastiehaiev:
I prepared PR with a fix. #215 
@seik Please review, I really need this fix for my bot.

zhtaihao:
@rastiehaiev You may be interested in #214, which also fixes this problem, in a different way than you did. I am currently using that approach and it is working well.

rastiehaiev:
@zhtaihao yeah, I saw it but decided to provide another approach. Approach I used is already currently used in the project, so I did the same (just converted media payload to JSON and store it as a string field). Hope new version will arrive with any of ours fixes so I can use it in my project 🙏.

rastiehaiev:
@zhtaihao my approach doesn't work. Declining PR...

zhtaihao:
This has been fixed in https://github.com/kotlin-telegram-bot/kotlin-telegram-bot/pull/214 @trupizzza 

Closing this issue now.

trupizzza:
> This has been fixed in #214 @trupizzza
> 
> Closing this issue now.

[...]

PR #214, opened 2021-08-05
JSON serialization of InputMedia Types

joshuastorch:
The InputMedia Classes were not serialized as JSON and instead printed as default data class strings. This implements a possible fix for that problem.
This Problem was reported in issue #213


vjgarciag96 (approved):
Thanks for the fix @joshuastorch! (and sorry for the late review)

LGTM. We usually create retrofit converters like https://github.com/kotlin-telegram-bot/kotlin-telegram-bot/blob/main/telegram/src/main/kotlin/com/github/kotlintelegrambot/network/retrofit/converters/ChatIdConverterFactory.kt rather than overriding the `toString` for serialisation purposes. But we can go ahead with the fix and move the change to a retrofit converter as a separated cleanup task.

joshuastorch:
I resolved the merge conflict.
I have little knowlege about retrofit and based the toString overrides on the ReplyMarkup Types like [InlineKeyboardMarkup](https://github.com/kotlin-telegram-bot/kotlin-telegram-bot/blob/main/telegram/src/main/kotlin/com/github/kotlintelegrambot/entities/InlineKeyboardMarkup.kt). 

I looked into the retrofit converters you mentioned but as far as i can see they are not called when using functions like [editMessageMedia](https://github.com/kotlin-telegram-bot/kotlin-telegram-bot/blob/main/telegram/src/main/kotlin/com/github/kotlintelegrambot/network/ApiClient.kt#L1062). If they were i think the [GsonConverter](https://github.com/kotlin-telegram-bot/kotlin-telegram-bot/blob/main/telegram/src/main/kotlin/com/github/kotlintelegrambot/network/ApiClient.kt#L135) would take care of it.

joshuastorch:
I added a new ```InputMediaConverterFactory``` and removed the ```toString``` methods. 
I had to add the ```TelegramFileAdapter``` and ```InputMediaAdapter``` to ```GsonFactory.createForApiClient()```

telegram/src/main/kotlin/com/github/kotlintelegrambot/entities/inputmedia/InputMedia.kt
```diff
@@ -76,8 +76,8 @@ data class InputMediaAnimation(
     @SerializedName(InputMediaFields.HEIGHT) val height: Int? = null,
     @SerializedName(InputMediaFields.DURATION) val duration: Int? = null
 ) : InputMedia() {
-    override val type: String
-        get() = InputMediaTypes.ANIMATION
+    @SerializedName(InputMediaFields.TYPE)
```
vjgarciag96, line 79: Nice one! I imagine we can now remove the code handling this manually in `InputMediaAdapter`?https://github.com/kotlin-telegram-bot/kotlin-telegram-bot/blob/main/telegram/src/main/kotlin/com/github/kotlintelegrambot/network/serialization/adapter/InputMediaAdapter.kt#L24-L25

vjgarciag96 (approved):
This looks great! Thank you so much for the contribution! Let's merge it

telegram/src/main/kotlin/com/github/kotlintelegrambot/entities/inputmedia/InputMedia.kt
joshuastorch: I think the entire custom serializer ```InputMediaAdapter``` for the ```InputMedia``` types can be removed.

PR #215, opened 2021-08-06
Fix serialization of InputMedia.

rastiehaiev:


rastiehaiev:
Provided a fix for #213

rastiehaiev:
Does not work as expected.

PR #222, opened 2021-10-05
Add missing parseMode parameter for editMessageCaption

joshuastorch:
ParseMode Parameter for [editMessageCaption](https://core.telegram.org/bots/api#editmessagecaption) was missing.


vjgarciag96 (approved):
LGTM! Thanks @joshuastorch

PR #223, opened 2021-10-06
Fix send media group

### LINKED ISSUE #221: ByteArray files support on sendMediaGroup
We don't support sending files as `ByteArray`s in the `sendMediaGroup` operation. We should add support for that.

A good point to start looking into this is https://github.com/kotlin-telegram-bot/kotlin-telegram-bot/blob/1773290206d217999c73a88bbeec824ba591b50f/telegram/src/main/kotlin/com/github/kotlintelegrambot/network/multipart/MultipartBodyFactory.kt.

naftalmm:
This fixes #221 and other issues related to sendMediaGroup method.

Also had to change the hierarchy of `InputMedia` and `GroupableMedia` classes/interfaces:

- Now they both are sealed interfaces
- Now `GroupableMedia` extends `InputMedia` (this better emphasize their relationship, since all data classes, which were extending `InputMedia` at the same time implemented `GroupableMedia` interface, and allows to access properties of `InputMedia` working with `GroupableMedia`)


telegram/src/main/kotlin/com/github/kotlintelegrambot/entities/inputmedia/InputMedia.kt
```diff
@@ -7,28 +7,55 @@ import com.google.gson.annotations.SerializedName
...
-            throw IllegalArgumentException("media groups must include 2-10 items")
-        }
+        require(medias.size in 2..10) { "media groups must include 2-10 items" }
     }
 
     companion object {
-        fun from(vararg media: GroupableMedia): MediaGroup = MediaGroup(media)
+        fun from(vararg media: GroupableMedia): MediaGroup = MediaGroup(generateUniqueNamesForNamelessByteArrays(media))
+
+        private fun generateUniqueNamesForNamelessByteArrays(media: Array<out GroupableMedia>): Array<GroupableMedia> {
```
vjgarciag96, line 31: I'm wondering if we should push the decision about the name for the `ByteArray` files to consumers of the lib. instead (i.e. making `filename` not optional in `TelegramFile.ByByteArray`). It feels like they are in a better position to make the right call here, and not forcing it in the API may lead to unexpected results (e.g. they may not know that if they don't put a name in their `ByByteArray` objects, a random `nonamex` or whatever will be used instead). And if they don't care about the name, they can use the default of of their choice (or we could even provide a default if we think that will be a common use case). What do you think?

telegram/src/main/kotlin/com/github/kotlintelegrambot/network/multipart/MultipartBodyFactory.kt
```diff
@@ -44,22 +41,21 @@ internal class MultipartBodyFactory(private val gson: Gson) {
...
-            groupableMedia is InputMediaVideo && groupableMedia.thumb != null -> listOf(
-                groupableMedia.thumb.file to MediaTypeConstants.IMAGE
-            )
-            else -> emptyList()
+    private fun MediaGroup.takeFilesParts(): List<MultipartBody.Part> = medias.flatMap { groupableMedia ->
+        val mediaType = when (groupableMedia) {
+            is InputMediaPhoto -> MediaTypeConstants.IMAGE
+            is InputMediaVideo -> MediaTypeConstants.VIDEO
+            is InputMediaAudio -> MediaTypeConstants.AUDIO
+            else -> null
```
vjgarciag96, line 49: Do you think it would be better to be exhaustive here and add all the possible branches here? That way, if a new `GroupableMedia` is added, this code won't compile and whoever is doing the changes will be forced to consider how the new type should be dealt with here.

telegram/src/main/kotlin/com/github/kotlintelegrambot/network/multipart/MultipartBodyFactory.kt
```diff
@@ -44,22 +41,21 @@ internal class MultipartBodyFactory(private val gson: Gson) {
...
         }
+        mutableListOf<MultipartBody.Part>().apply {
+            when (val media = groupableMedia.media) {
+                is TelegramFile.ByFile -> add(media.file.toMultipartBodyPart(mediaType = mediaType))
+                is TelegramFile.ByByteArray -> add(media.fileBytes.toMultipartBodyPart(partName = media.filename!!, mediaType = mediaType))
+            }
+            if (groupableMedia is InputMediaVideo && groupableMedia.thumb != null) {
+                add(groupableMedia.thumb.file.toMultipartBodyPart(mediaType = MediaTypeConstants.IMAGE))
+            }
+        }.toList()
```
vjgarciag96, line 59: Not important, but just out of curiosity sharing the `buildList` function (https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.collections/build-list.html) that can be very handy for cases like this.

vjgarciag96 (commented):
Getting very late to this one, sorry @naftalmm. Thank you so much for the contribution, I love it! I left a couple of small comments.

telegram/src/main/kotlin/com/github/kotlintelegrambot/entities/inputmedia/InputMedia.kt
naftalmm: Making `filename` not optional will be a breaking change. So, if we want to make it non-nullable, we still need to provide some default (but now non-null) value (strictly speaking, it will be breaking change anyway, but only for someone who explicitly passed `null` for this parameter, repeating current default).
Not sure why I thought that Telegram API requires filenames uniqueness across `MediaGroup`. Revised their docs, but haven't found it. So, yes, we could provide some non-null default in TelegramFile.ByByteArray constructor (and not worry about filename collision if there will be two `TelegramFile`s with the same default value in `MediaGroup`) - it will greatly simlify the code and indeed will make interaction with API more straightforward.

While revising docs for [sendMediaGroup](https://core.telegram.org/bots/api#sendmediagroup) found another restriction, we're missing check for: 

>Documents and audio files can be only grouped in an album with messages of the same type.

Will implement it the scope of this PR, because this very thing happens in the tests I've added. 🙈🙈🙈

telegram/src/main/kotlin/com/github/kotlintelegrambot/network/multipart/MultipartBodyFactory.kt
naftalmm: Good point. Fixed.

telegram/src/main/kotlin/com/github/kotlintelegrambot/network/multipart/MultipartBodyFactory.kt
naftalmm: Indeed, `buildList` perfectly suits here, but first we need to bump Kotlin version at least to 1.6. Otherwise, we'll have to mess with `@ExperimentalStdlibApi` annotations

seik:
Hi @naftalmm — first of all, thanks a lot for this PR and the great writeup back in 2021, it really helped frame the problem. After 4 years of inactivity and the diff drifting away from main, we've taken a more incremental path: #335 just landed which fixes the core ByteArray-in-mediaGroup issue (closes #221) without the wider sealed-interface refactor you proposed here.

The sealed-interface cleanup is still a good idea on its own — happy to revisit it in a follow-up if anyone wants to pick it up against current main. Closing this for now. Thanks again!

PR #224, opened 2021-10-08
Update Gradle to 7.2

naftalmm:
- Fixed all deprecation warnings
- Removed Gradle duplicate from `telegram` subproject (manual removal of `telegram/.gradle` folder may be required)
- Updated Ktor dependency, so that all Kotlin stdlib versions (from Gradle, from Kotlin plugin. from Ktor) were aligned


seik (approved):
A bit out of touch with kotlin/jvm ecosystem but looks good. Thanks for contributing!

PR #235, opened 2021-10-30
dice_roller_app.kt

Yash-passi:


vjgarciag96:
Hi @Yash-passi! This doesn't have anything to do with the Kotlin Telegram Bot library, can you elaborate on what your contribution is about?

Issue #158, opened 2021-01-30
Add support for live location heading


vjgarciag96:
In Telegram Bot Api 5.0, support for live location [heading](https://en.wikipedia.org/wiki/Heading_(navigation)) was added:

* Added field `heading` to `Location`, `InlineQueryResultLocation` and `InputLocationMessageContent`.
* Added parameter `heading` to the methods `sendLocation` and `editMessageLiveLocation`.

https://core.telegram.org/bots/api#location
https://core.telegram.org/bots/api#inlinequeryresultlocation
https://core.telegram.org/bots/api#inputlocationmessagecontent
https://core.telegram.org/bots/api#sendlocation
https://core.telegram.org/bots/api#editmessagelivelocation

We should implement this support.

PR #247, opened 2022-06-14
feat: Added location heading field

joaoGabriel55:
Fixed issue #158

### DIFF
--- a/telegram/src/test/kotlin/com/github/kotlintelegrambot/entities/LocationTest.kt
+++ b/telegram/src/test/kotlin/com/github/kotlintelegrambot/entities/LocationTest.kt
@@ -1,0 +1,19 @@
+package com.github.kotlintelegrambot.entities
+
+import org.junit.jupiter.api.Assertions
+import org.junit.jupiter.api.Test
+
+internal class LocationTest {
+    @Test
+    fun `should permit to not inform heading optional attribute value`() {
+        val location = Location(23.4F, 45.0F)
+        Assertions.assertEquals(location.heading, null)
+    }
+
+    @Test
+    fun `should permit to inform heading optional attribute value`() {
+        val locationHeading = 34
+        val location = Location(23.4F, 45.0F, locationHeading)
+        Assertions.assertEquals(location.heading, locationHeading)
+    }
+}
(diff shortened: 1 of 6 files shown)


joaoGabriel55:
Two questions @vjgarciag96: 
- I could not found `InlineQueryResultLocation` and `InputLocationMessageContent` files/classes. Do they still exist?
- How can I create test suites for `sendLocation` and `editMessageLiveLocation` API client methods?

telegram/src/test/kotlin/com/github/kotlintelegrambot/entities/LocationTest.kt
```diff
@@ -0,0 +1,19 @@
...
+        val location = Location(23.4F, 45.0F)
+        Assertions.assertEquals(location.heading, null)
+    }
+
+    @Test
+    fun `should permit to inform heading optional attribute value`() {
+        val locationHeading = 34
+        val location = Location(23.4F, 45.0F, locationHeading)
+        Assertions.assertEquals(location.heading, locationHeading)
+    }
```
vjgarciag96, line 18: No need to change, but it feels like we don't need this test - it's mainly testing a Kotlin language feature rather than actual logic of the library.

vjgarciag96 (commented):
Looks good so far!

Re. your questions:

* `InlineQueryResultLocation` can be found in https://github.com/kotlin-telegram-bot/kotlin-telegram-bot/blob/ff3e1b7da9aa8470a717190ef90964e97eca7bc3/telegram/src/main/kotlin/com/github/kotlintelegrambot/entities/inlinequeryresults/InlineQueryResult.kt#L158
* `InputLocationMessageContent` can be found in https://github.com/kotlin-telegram-bot/kotlin-telegram-bot/blob/ff3e1b7da9aa8470a717190ef90964e97eca7bc3/telegram/src/main/kotlin/com/github/kotlintelegrambot/entities/inlinequeryresults/InputMessageContent.kt#L13
* As for the tests, we could have something similar to what we have for other API operation https://github.com/kotlin-telegram-bot/kotlin-telegram-bot/blob/ff3e1b7da9aa8470a717190ef90964e97eca7bc3/telegram/src/test/kotlin/com/github/kotlintelegrambot/network/apiclient/SendDiceIT.kt#L14

seik:
Hi @joaoGabriel55 — thanks for the contribution! The branch was too far behind main to rebase cleanly after 3 years, so I've reimplemented the `heading` field directly on main (52cf7e8) together with the `horizontal_accuracy` work from #248. You're listed as co-author on the commit and the commit closes #158. Really appreciate the original work — closing here.

PR #258, opened 2022-09-04
fix issue #241

### LINKED ISSUE #241: edited_message not handled
**Expected behavior**
edited_message must be handled.

**Actual behavior**
edited_message not handled.

**To Reproduce**
just edit an existing message.
dispatch { message {} } won't work.

**Fast solution**

[Added_handling_edited_message_patch.txt](https://github.com/kotlin-telegram-bot/kotlin-telegram-bot/files/8496230/Added_handling_edited_message_patch.txt)

Takexito:
close #241

### DIFF
--- a/telegram/api/telegram.api
+++ b/telegram/api/telegram.api
@@ -360,17 +360,19 @@ public final class com/github/kotlintelegrambot/dispatcher/handlers/LocationHand
 }
 
 public final class com/github/kotlintelegrambot/dispatcher/handlers/MessageHandlerEnvironment {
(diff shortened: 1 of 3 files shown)


vjgarciag96 (commented):
Changes look good! What I wonder is whether consumers of the library will expect to receive message "edits" in their message handlers. I wonder if having a separate `messageEdited` handler would be better? Or offering an option to customise whether you want to receive message edits in the message handler? What do you think? Thanks for your contribution btw!

seik:
Hi @Takexito — thanks a lot for the PR and apologies for the very long wait. We ended up going with #355, which takes the approach @vjgarciag96 suggested in your review (a separate `textEdit` handler rather than mixing edits into the existing message handler). That PR is now merged.

Really appreciate the contribution, sorry it didn't make it in directly. Closing.

PR #261, opened 2022-10-08
Add close method

### LINKED ISSUE #147: Add close method
A new `close` method was added in Telegram Bot Api 5.0:

>Use this method to close the bot instance before moving it from one local server to another. You need to delete the webhook before calling this method to ensure that the bot isn't launched again after server restart. The method will return error 429 in the first 10 minutes after the bot is launched. Returns True on success. Requires no parameters.

https://core.telegram.org/bots/api#close

We should add support for this method.

lpicanco:
Resolves #147.

Let me know if you need any adjustments.


lpicanco:
@vjgarciag96, updated the `telegram.api` file. 
Can you run the workflow again?

vjgarciag96 (approved):
Thanks for your contribution @lpicanco! Looks good to me

Issue #262, opened 2022-10-10
How to call fun removeHandler(handler: Handler) correctly


lifestreamy:
Say, for example, I have many nested callback queries, and they are added to bot handlers as they are created

```
command("enterdata") {
                        val data = mutableListOf<String>()
                        bot.sendMessage(
                            ChatId.fromId(update.message!!.chat.id),
                            text = "asdsad",
                            replyMarkup = InlineKeyboardMarkup.create(
                                listOf(InlineKeyboardButton.CallbackData(text = "data1", callbackData = "data1"))
                            )
                        )
                        var callbackText1 = "data1"
                        var callbackText2 = "data2"
                        var callbackText3 = "data3"
                        callbackQuery(callbackText1) callback1@{
                            val chatId1 = callbackQuery.message?.chat?.id ?: return@callback1
                            if (callbackQuery.data == callbackText1) {
                                bot.sendMessage(
                                    chatId = ChatId.fromId(chatId1),
                                    text = "received cb data1",
                                    replyMarkup = InlineKeyboardMarkup.create(
                                        listOf(
                                            InlineKeyboardButton.CallbackData(
                                                text = "data2",
                                                callbackData = callbackText2
                                            )
                                        )
                                    )
                                )
                            }
//                            callbackText1 = null.toString()
                            callbackQuery(callbackText2) callback2@{
                                val chatId2 = callbackQuery.message?.chat?.id ?: return@callback2
                                if (callbackQuery.data == callbackText2) {
                                    bot.sendMessage(
                                        chatId = ChatId.fromId(chatId2),
                                        text = "received cb data2",
                                        replyMarkup = InlineKeyboardMarkup.create(
                                            listOf(
                                                InlineKeyboardButton.CallbackData(
                                                    text = "data3",
                                                    callbackData = callbackText3
                                                )
                                            )
                                        )
                                    )
                                }
//                                callbackText2 = null.toString()
                                callbackQuery(callbackText3) callback3@{
                                    val chatId3 = callbackQuery.message?.chat?.id ?: return@callback3
                                    if (callbackQuery.data == callbackText3) {
                                        bot.sendMessage(
                                            chatId = ChatId.fromId(chatId3),
                                            text = "received cb data3"
                                        )
                                    }
//                                    callbackText3 = null.toString()
                                }
                            }
                        }
                    }
```

[...]

PR #267, opened 2022-10-18
Changed addHandler() and removeHandler()

lifestreamy:
Changed HandlersDsl.kt to account for changes in addHandler

You can now add handlers in dispatcher specifying their names, every handler is added in a mutableMap with its name as a key. If no name is specified then it is either defaulted to null or to main parameter of handler

### DIFF
--- a/README.md
+++ b/README.md
@@ -4,6 +4,22 @@
 [![ktlint](https://img.shields.io/badge/code%20style-%E2%9D%A4-FF4081.svg)](https://ktlint.github.io/)
 
 A wrapper for the Telegram Bot API written in Kotlin.
+```
+**This fork was modified by @lifestreamy (me) for my own purposes.**
+
+Out of 302 tests from the original repository the following 3 have not been passed (due to the changing of some core behaviour) and thus have been skipped:
+-DispatcherTest. test that handlers from different groups are called in consistent order()
+-SetWebhookIT. setWebhook with certificate as file and with ip address$telegram()
+-SetWebhookIT. setWebhook with certificate as file$telegram()
+
+The main changes introduced to this fork are:
+-You can now add handlers specifying their names to identify them, addHandler(... , handlerName = "some_name" , ...)
+-You can now remove handlers with removeHandler("handler_name")
+-You can now add handlers to specific chats, addHandler(... , chatId = *pass chat id here* , ...)
+
+Be warned: the stability of this fork is not yet guaranteed and is to be tested in actual application
+```
+
 
 ## Usage
 
--- a/samples/dispatcher/src/main/kotlin/com/github/kotlintelegrambot/dispatcher/Main.kt
+++ b/samples/dispatcher/src/main/kotlin/com/github/kotlintelegrambot/dispatcher/Main.kt
@@ -49,12 +49,27 @@ fun main() {
                     }
                 )
             }
+
             /**
              * Testing removing commands
              */
             command("remove_this_command") {
                 removeHandler("/remove_this_command")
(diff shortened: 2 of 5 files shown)


lifestreamy:
This commit is made to address this issue that I have posted recently
https://github.com/kotlin-telegram-bot/kotlin-telegram-bot/issues/262

vjgarciag96:
Thanks for the contribution @lifestreamy! This will take me some time to review as I'll have to thoroughly go through your  proposal and think about it because it's changing a very core part of the library and I'm not sure this is the behaviour we want, and in case we want it, I'll need to check whether this is the way we'd want to implement it. I'll add a suggestion on your issue on how you could solve the issue without changing the library in the meantime if that works for you.

lifestreamy:
Alright. Thank you for the answer.

lifestreamy:
And yes, it probably is not the best way to solve the problem and it could be improved. But it works for me for now, although I don't really like some pieces in my code.

lifestreamy:
And in addHandler I added redundant check for equality of class names, it is not not needed since in a map there is only 1 value for each key.

So yes, it definitely needs improvements, but I think you've got my point.

lifestreamy:
Some clarification: in addHandler() it is written so that it does not replace existing handler with the same one (same id and type). If types are different it replaces the handler previously added under the same id

lifestreamy:
Also it would be cool to be able to add 
a nested handler like this

callbackQuery("cb1", chatId){}

This way it will be only added to the selected chat


And, to avoid cluttering, I think it is alright to restart the backend periodically,
but it is probably possible (and would be better) to improve the library so that you could just reset the bot every select interval.

Also, there is a problem with dispatcher that you can't just send message every time the bot starts or on some backend event (like when there is an update a message will be sent to every person using the bot with the news)

Currently I have to manually send such messages using telegram API directly.

Like when the bot is started, it sends only to my own chat "Backend started"

I can't just write bot.sendMessage in dispatcher, it is not allowed by the library.

Right now the bot can only send messages from inside a handler block.

vjgarciag96:
I take note of all the feedback. The suggestion around making it easier to add/remove handlers, and adding the same handler several times will be in the roadmap for new features of the library. The handler per chat id, we'll have to give that a thought and see whether we want that or not. And the ability to reset the bot sounds like a good idea too.

> Also, there is a problem with dispatcher that you can't just send message every time the bot starts or on some backend event (like when there is an update a message will be sent to every person using the bot with the news)

> Currently I have to manually send such messages using telegram API directly.

> Like when the bot is started, it sends only to my own chat "Backend started"

> I can't just write bot.sendMessage in dispatcher, it is not allowed by the library.

> Right now the bot can only send messages from inside a handler block.

The `dispatcher` block is not the best place where you want to use the bot as it should only be used to define handlers for updates or errors. If you want to send a message when you start a bot you can do something like the next:

```kotlin
fun main() {
    val bot = bot {
        dispatcher {
            // whatever goes here
        }    
    }
    bot.startPolling()
    bot.sendMessage(ANY_CHAT_ID, ANY_MESSAGE)
}
```

lifestreamy:
Thank you for your replies! It is great that you answer and care about your product, keep up the good work! Meanwhile I will use my fork. And yes, I haven't run any tests before making pull request. Whoops lol

lifestreamy:
https://github.com/lifestreamy/kotlin-telegram-bot/blob/main/README.md
Please, check this out.

Again, just to get some attention to the issue. 

I can (and will) delete this pull request on your demand.

PR #277, opened 2023-02-28
Use coroutines

### LINKED ISSUE #251: Unable to launch coroutines inside of the dispatch block
**Expected behavior**
Coroutine should launch.

**Actual behavior**
Coroutine doesn't launch.

**To Reproduce**
Make the `main` function `runBlocking` and then, try to launch new coroutines using `launch { }` inside the `dispatcher { message { } }` block.

Maybe I'm just doing something wrong...

JcMinarro:
Convert `Handler.handleUpdate()` method into a suspend method. It allows calling coroutines method from dispatcher block code.
The `ExecutorLooper` use coroutines as well instead of a `ThreadPoolExecutor` that only runs a new thread.
Fix #251


vjgarciag96:
> @seik @vjgarciag96 are you keeping maintaining this repo?
> Could I have a review of this PR?

It's still maintained, but as you can see it has not been very active lately. I'll try and have a look at the code through the weekend. Thank you so much for the contribution, excited to finally see coroutines support in the library!

JcMinarro:
Cool!!!
Thank you for the project, it is handy.
If you feel something needs to be changed, let me know 🙃

telegram/src/main/kotlin/com/github/kotlintelegrambot/dispatcher/Dispatcher.kt
```diff
@@ -7,35 +7,40 @@ import com.github.kotlintelegrambot.entities.Update
...
+import kotlinx.coroutines.launch
+import kotlinx.coroutines.yield
 import java.util.concurrent.BlockingQueue
-import java.util.concurrent.Executor
 
 class Dispatcher internal constructor(
     private val updatesQueue: BlockingQueue<DispatchableObject>,
-    private val updatesExecutor: Executor,
     private val logLevel: LogLevel,
+    ioDispatcher: CoroutineDispatcher = Dispatchers.IO,
```
vjgarciag96, line 21: I wonder if we should be using `Dispatchers.IO` here. The dispatcher here is used to launch a single coroutine that will loop until stopped, checking for updates on a queue, and running functions provided by consumers of the library for each of those (in those functions consumers can decide whether they need to run in a different dispatcher or not). But we don't know whether those functions will be blocking IO operations so it doesn't feel right to use this `Dispatcher` here. The main downside I could see of this is reducing the potential for parallelism of `Dispatchers.IO`, which is global to the app, because we have one of its threads always busy. In practice, I'm not sure whether this will actually make a difference but it may for some use cases? Any thoughts on this? One alternative could be to just have a dispatcher backed by a `singleThreadExecutor` that we create only for this.

A bit unrelated. This also makes me wonder about next steps for better coroutine adoption, and feels like replacing `BlockingQueue` for a `Channel` could be a nice impovement, so we suspend while waiting for new updates rather than blocking. Also out of scope of this PR, but just bringing it up to get your thoughts about it.

telegram/src/main/kotlin/com/github/kotlintelegrambot/updater/Looper.kt
```diff
@@ -12,24 +17,23 @@ internal interface Looper {
  * intended to run the loop in a different thread). The loop will stop if the thread running the
  * loop is interrupted or in the next iteration after the [quit] method is called.
  */
-internal class ExecutorLooper(
-    private val loopExecutor: Executor,
-) : Looper {
+internal class ExecutorLooper(ioDispatcher: CoroutineDispatcher) : Looper {
```
vjgarciag96, line 20: Not important. The name of this class and the comment seems a bit outdated now that we are using a `CoroutineDispatcher` rather than an `Executor`.

telegram/src/main/kotlin/com/github/kotlintelegrambot/updater/Looper.kt
```diff
@@ -12,24 +17,23 @@ internal interface Looper {
...
 
-    @Volatile private var isLooping = false
+    private val scope: CoroutineScope = CoroutineScope(ioDispatcher)
+    private var job: Job? = null
```
vjgarciag96, line 23: Similar comments to the ones in the `Dispatcher` class:

* We may want to make `job` volatile for memory consistency because we can't guarantee `loop` and `quit` are going to be called from the same thead, and both are accessing `job`.
* I also wonder about the `Dispatcher` we should be using here. In this case, this is actually used to run the API operations so it might make sense for it to be `Dispatchers.IO`.

telegram/src/test/kotlin/com/github/kotlintelegrambot/updater/BoundLooper.kt
```diff
@@ -1,15 +1,22 @@
...
+class BoundLooper(coroutineDispatcher: CoroutineDispatcher) : Looper {
 
+    private val scope: CoroutineScope = CoroutineScope(coroutineDispatcher)
     var loopIterations = 0
 
-    override fun loop(loopBody: () -> Unit) {
-        repeat(loopIterations) {
-            loopBody()
+    override fun loop(loopBody: suspend () -> Unit) {
+        scope.launch {
```
vjgarciag96, line 16: Is there a reason why we need to launch a coroutine here for some of the tests or could we keep this being executed in the caller context?

telegram/src/test/kotlin/com/github/kotlintelegrambot/updater/ExecutorLooperTest.kt
```diff
@@ -1,61 +1,35 @@
...
-            }
-        }
-
-        assertEquals(13, count)
-    }
-
-    @Test
-    fun `loops until thread is interrupted`() {
-        var count = 0
+        println("JcLog: Start test $expectedCount")
```
vjgarciag96, line 22: I imagine this test was giving some problems with the new changes

telegram/src/test/kotlin/com/github/kotlintelegrambot/updater/ExecutorLooperTest.kt
```diff
@@ -1,61 +1,35 @@
...
+                sut.quit()
             }
         }
+        advanceUntilIdle()
 
-        assertEquals(13, count)
-    }
-
-    @Test
-    fun `loops until an exception is thrown`() {
```
vjgarciag96, line 45: Is this one not needed anymore? What happens when a `Exception` is thrown now?

vjgarciag96 (commented):
This is looking great! Thank you so much for the work. I added a few small comments, but nothing major. I think this version is very close to be ready to merge and ship 🚀

telegram/src/main/kotlin/com/github/kotlintelegrambot/dispatcher/Dispatcher.kt
JcMinarro: `Dispatcher` is a bit different than `Thread`, A `Dispatcher` is more like a `ThreadPool` where coroutines are executed (Except the `Dispatcher.Main` on Android implementation that use the UI Thread`), so it shouldn`t be blocking IO Operations on the rest of the app. Even the coroutine we launch on this Dispatcher will be running indefinitely until the loop is stopped, on every iteration of our infinite loop we are "suspending" our logic and delegating to the dispatcher we can continue (our job has not been canceled) or if there are others coroutines waiting to be executed and need to be run before us. We achieve it by calling `yield()` method.

Apart from that, I had a doubt about using `Dispatchers.Default` or `Dispatchers.IO`. Reading the [docs](https://kotlinlang.org/api/kotlinx.coroutines/kotlinx-coroutines-core/kotlinx.coroutines/-coroutine-dispatcher/) First one is used to run "hard jobs" that consume machine recurses and the second one to operations that will be interacting with IO Resources. In our case, I think the IO definition is closer to our implementation (We run IO calls on every iteration). Maybe, something we could do is run `updates` on `Dispatchers.Default` but switch to `Dispatchers.IO` when API calls are executed. For that would be better to improve the "Retrofit integration" to use coroutines internally, but I think it can be achieved in a separate PR if we decide it is needed.

Lastly, regarding the `BlockingQueue`, of course, we could use `Channel` instead, I didn't think about that, but it is a really good point

telegram/src/main/kotlin/com/github/kotlintelegrambot/updater/Looper.kt
JcMinarro: I am thinking and, as far as we cancel the job just before a new job is started, we shouldn't have memory problems. A job can be called to `cancel()` multiple times.

telegram/src/test/kotlin/com/github/kotlintelegrambot/updater/ExecutorLooperTest.kt
JcMinarro: When an exception is thrown, the job is stopped, so maybe we can keep this test as well, but I don't think it gives us a lot of value.
Let me see if I can keep it as well

telegram/src/main/kotlin/com/github/kotlintelegrambot/dispatcher/Dispatcher.kt
JcMinarro: I will try to do separate PR using `Channel` instead of `BlockingQueue`, because If I change it here, lot of code will need to be updated (Tests are using fake blockingQueue that will need to be updated as well..)

JcMinarro (commented):
Thanks for the review.
Let me update the PR

telegram/src/main/kotlin/com/github/kotlintelegrambot/dispatcher/Dispatcher.kt
vjgarciag96: Yes, right. I see the point of this code being "cooperative" by doing `yield` every time an item is received from the queue, so it's not going to be taking a thread forever. For now, it will only be blocking in `queue.take` but that will become suspending when we migrate to use `Channel` so I think this class is doing good.

But I don't think we run any IO operation in `Dispatcher` (we run them on `Updater`) so that's why I was thinking that maybe it would be nicer to use our own dispatcher backed by a single thread executor so that we don't take any of the resources from `Dispatchers.IO` here. There's this nice extension function to make our work easy https://kotlinlang.org/api/kotlinx.coroutines/kotlinx-coroutines-core/kotlinx.coroutines/as-coroutine-dispatcher.html.

I agree it could make sense to use `Dispatchers.IO` on the `Updater` class, which is where we actually run the `getUpdates` API operation.

Re. Retrofit it might be nice to use the suspend version of the functions but as far as I understand OkHttp doesn't support non-blocking IO operations yet (https://github.com/square/okio/pull/531). One alternative would be to use Ktor instead https://github.com/ktorio/ktor, but that would be a big task, so I guess we can do with execution blocking IO operations on `Dispatchers.IO` for now.

telegram/src/main/kotlin/com/github/kotlintelegrambot/updater/Looper.kt
vjgarciag96: The memory consistency point was about memory visibility, sorry I may have made it confusing.

As we don't know or control what thread a consumer is calling `loop` and `quit` from, we could have thread A calling `loop` and thread B calling `quit`. `loop` writes `job` and `quit` reads it, and without additional synchronization, it's not guaranteed that thread B will be able to see the value that thread A wrote to `job`. That would be solved by making `job` volatile. Does this make sense?

telegram/src/test/kotlin/com/github/kotlintelegrambot/updater/BoundLooper.kt
vjgarciag96: Ah true, that makes sense. Thanks for the explanation.

telegram/src/test/kotlin/com/github/kotlintelegrambot/updater/ExecutorLooperTest.kt
vjgarciag96: Ahh I see, but the exception is still propagated because we are using `launch`?

vjgarciag96 (commented):
Great stuff, thanks for actioning the feedback! Once we resolve the discussion around the `Dispatcher` and the volatile we are good to go.

telegram/src/main/kotlin/com/github/kotlintelegrambot/dispatcher/Dispatcher.kt
JcMinarro: Ok, I see. Let me create our own `Dispatcher` and inject it, instead of injecting the `Dispatcher.IO` 

Re Network Layer, I think it is out of the scope of this PR, it can be covered later

telegram/src/main/kotlin/com/github/kotlintelegrambot/updater/Looper.kt
JcMinarro: Ok, make sense. Let me update it.

telegram/src/main/kotlin/com/github/kotlintelegrambot/dispatcher/Dispatcher.kt
vjgarciag96: Yes, completely agree.

vjgarciag96 (approved):
LGTM! This is wonderful, thank you so much for the contribution again!

JinKoro:
@vjgarciag96 When is the release of this task planned?

JcMinarro:
> @vjgarciag96 When is the release of this task planned?

I am not sure if there is any new release planned yet.
As far as this is on Jitpack, you can use a "commit SHA1" as a dependency version in your project.
```
dependencies {
  implementation 'com.github.kotlin-telegram-bot.kotlin-telegram-bot:dispatcher:f2c594d16c'
}
```

PR #280, opened 2023-03-10
Add my_chat_member to update

Zheg-al:
Add my_chat_member  to update
according to https://core.telegram.org/bots/api#update


JcMinarro:
> Add my_chat_member to update according to https://core.telegram.org/bots/api#update

Hello  @Zheg-al 
Thanks for the PR.
Looks like this PR contains multiple features (the new field `my_chat_member`, an option to handle exceptions, ...)
Could you split the PR into different ones?

seik:
Hi @Zheg-al — thanks for the PR! In the meantime #301 landed and already covers `my_chat_member` on `Update` together with a `MyChatMemberHandler` and the `myChatMember { ... }` dispatcher DSL (commit ea90d82). Both are on `main` today, so the core of your change is already in.

If there's a piece of your original PR that isn't covered by #301 (e.g. the exception-handling option), please feel free to open a focused PR for that piece against current `main`. Closing this for now. Really appreciate the contribution!

PR #284, opened 2023-05-13
Upgrade ktlint

JcMinarro:
Upgrade KtLint version to a version that requires Traling Commas to avoid PR conflicts


.git-blame-ignore-revs
```diff
@@ -0,0 +1,2 @@
+# Apply new format to the code
+c8ef601925956c98b052034cb0f38cffc88e0aae
```
JcMinarro, line 2: This file allows us to skip any commit from git-blame that doesn't apport anything to the history.
In that case, this commit is the one where the check style has been updated, but the code has no feature change.
[Here](https://docs.github.com/en/repositories/working-with-files/using-files/viewing-a-file#ignore-commits-in-the-blame-view) you can read docs about it.

We shouldn't squash this PR to be sure the commit SHA1 doesn't change

vjgarciag96 (approved):
Nice, thank you!

vjgarciag96:
I think this now has conflicts with the PR to open handlers that got merged just now.

JcMinarro:
I will update it

vjgarciag96 (approved):

PR #287, opened 2023-05-14
Open Handler classes

### LINKED ISSUE #286: Open `Handler` Classes to be able to extend them
**Describe the solution you'd like**
Handlers Implementations are final internal classes and don't allow customizations.
Could be interesting to add extra customization to those classes. Imagine I want to create a custom `CommandHandler` that apart of check if user called the `/command`, contains a valid argument. If I could extend the `CommandHandler` class it would be as easy as:
```
val customCommand = object : CommandHandler("command", {
    // Logic to consume the update
}) {
    override fun checkUpdate(update: Update): Boolean =
        super.checkUpdate(update) &&
            update.message?.text?.contains("light") ?: false
}
```

**Implementation ideas**
Declare `Handler`'s implementation as `public open class`

JcMinarro:
Fix #286


vjgarciag96 (approved):
Looks great, thanks!

PR #290, opened 2023-06-18
Create InputChainHandler 

JcMinarro:
This new `InputChainHandler` helps us to collect inputs after a previous handler has been processed.

Let show some examples:

-. A button is pressed to change the name of the device:
```
addHandler(
        CallbackQueryHandler(
            callbackData = "change-device-name",
            handleCallbackQuery = {
                bot.sendMessage(
                    chatId = ChatId.fromId(callbackQuery.from.id),
                    text = "Which name do you want to use?"
                )
                update.consume()
            }
        ).inputChain<String>(
            inputChainIdGen = { previousInputChainId, update -> "someId" },
            handleInputChain = {
                val name = this.text
                bot.sendMessage(
                    chatId = ChatId.fromId(message.from?.id!!),
                    text = "Ok, I will use '$name' for your device"
                )
                update.consume()
                // Return `null` if we don't want to receive new updates on the chain
                null
            }
        )
    )
```

-. The user send a command to create a new form
```
addHandler(
        CommandHandler("newForm") {
            bot.sendMessage(
                chatId = ChatId.fromId(message.from?.id!!),
                text = "The form will have 3 questions.\n1º What is your name?"
            )
            update.consume()
        }.inputChain<Int>(
            inputChainIdGen = { previousInputChainId, update ->
                         when {
                             previousInputChainId == null -> 0
                             previousInputChainId < 2 -> previousInputChainId + 1
                             else -> null
                         }
                              },
            handleInputChain = {
                when(inputChainId) {
                    0 -> {
                        val userName = text
                        bot.sendMessage(
                            chatId = ChatId.fromId(message.from?.id!!),
                            text = "Hello $userName, what month were you born?"
                        )
                    }
                    1 -> {
                        val month = text
                        bot.sendMessage(
                            chatId = ChatId.fromId(message.from?.id!!),
                            text = "$month, what year?"
                        )
                    }
                    2 -> {
                        val year = text
                        bot.sendMessage(
                            chatId = ChatId.fromId(message.from?.id!!),
                            text = "Cool, you have completed the form"
                        )
                    }
                }
                update.consume()
                inputChainIdGen(inputChainId, update)
            }
        )
    )
```

### DIFF
--- a/telegram/src/main/kotlin/com/github/kotlintelegrambot/dispatcher/HandlersDsl.kt
+++ b/telegram/src/main/kotlin/com/github/kotlintelegrambot/dispatcher/HandlersDsl.kt
@@ -193,8 +193,9 @@ fun Dispatcher.chatMember(handleChatMember: HandleChatMember) {
 }
 
 fun <T> Handler.inputChain(
+    filter: Filter = Filter.Text,
     inputChainIdGen: (previousInputChainId: T?, Update) -> T?,
     handleInputChain: HandleInputChain<T>,
 ): Handler {
-    return InputChainHandler(this, inputChainIdGen, handleInputChain)
+    return InputChainHandler(filter, this, inputChainIdGen, handleInputChain)
 }
--- a/telegram/src/main/kotlin/com/github/kotlintelegrambot/dispatcher/HandlersDsl.kt
+++ b/telegram/src/main/kotlin/com/github/kotlintelegrambot/dispatcher/HandlersDsl.kt
@@ -20,6 +20,7 @@ import com.github.kotlintelegrambot.dispatcher.handlers.HandleDocument
 import com.github.kotlintelegrambot.dispatcher.handlers.HandleError
 import com.github.kotlintelegrambot.dispatcher.handlers.HandleGame
 import com.github.kotlintelegrambot.dispatcher.handlers.HandleInlineQuery
+import com.github.kotlintelegrambot.dispatcher.handlers.HandleInputChain
 import com.github.kotlintelegrambot.dispatcher.handlers.HandleLeftChatMember
 import com.github.kotlintelegrambot.dispatcher.handlers.HandleLocation
 import com.github.kotlintelegrambot.dispatcher.handlers.HandleMessage
@@ -36,6 +37,7 @@ import com.github.kotlintelegrambot.dispatcher.handlers.HandleVideoNote
 import com.github.kotlintelegrambot.dispatcher.handlers.HandleVoice
 import com.github.kotlintelegrambot.dispatcher.handlers.Handler
 import com.github.kotlintelegrambot.dispatcher.handlers.InlineQueryHandler
+import com.github.kotlintelegrambot.dispatcher.handlers.InputChainHandler
 import com.github.kotlintelegrambot.dispatcher.handlers.LeftChatMemberHandler
 import com.github.kotlintelegrambot.dispatcher.handlers.LocationHandler
 import com.github.kotlintelegrambot.dispatcher.handlers.MessageHandler
@@ -188,4 +190,11 @@ fun Dispatcher.myChatMember(handleMyChatMember: HandleMyChatMember) {
 
 fun Dispatcher.chatMember(handleChatMember: HandleChatMember) {
     addHandler(ChatMemberHandler(handleChatMember = handleChatMember))
+}
+
+fun <T> Handler.inputChain(
+    inputChainIdGen: (previousInputChainId: T?, Update) -> T?,
+    handleInputChain: HandleInputChain<T>,
+): Handler {
+    return InputChainHandler(this, inputChainIdGen, handleInputChain)
 }
--- a/telegram/src/main/kotlin/com/github/kotlintelegrambot/dispatcher/handlers/InputChainHandler.kt
+++ b/telegram/src/main/kotlin/com/github/kotlintelegrambot/dispatcher/handlers/InputChainHandler.kt
@@ -1,0 +1,69 @@
+package com.github.kotlintelegrambot.dispatcher.handlers
+
+import com.github.kotlintelegrambot.Bot
+import com.github.kotlintelegrambot.entities.Message
+import com.github.kotlintelegrambot.entities.Update
+
+data class InputChainHandlerEnvironment<T>(
+    val inputChainIdGen: (previousInputChainId: T?, Update) -> T?,
+    val inputChainId: T,
+    val bot: Bot,
+    val update: Update,
+    val message: Message,
+    val text: String,
+)
+
+class InputChainHandler<T>(
+    private val initialHandler: Handler,
+    private val inputChainIdGen: (previousInputChainId: T?, Update) -> T?,
+    private val handleInputChain: HandleInputChain<T>,
+) : Handler {
+    private val chainMap = mutableMapOf<Long, T?>()
+
+    override fun checkUpdate(update: Update): Boolean =
+        initialHandler.checkUpdate(update) || update.check()
+
+    override suspend fun handleUpdate(bot: Bot, update: Update) {
+        update.message?.let { message ->
+            message.from?.id?.let { userId ->
+                chainMap.remove(userId)?.let { inputChainId ->
+                    message.text
+                        ?.takeUnless { it.isBlank() }
+                        ?.let { text ->
+                            handleInputChain(
+                                InputChainHandlerEnvironment(
+                                    inputChainIdGen,
+                                    inputChainId,
+                                    bot,
+                                    update,
+                                    message,
+                                    text,
+                                ),
+                            )?.let { chainMap[userId] = it }
+                        }
+                    Unit
+                }
+            }
+        } ?: initialHandler.handleUpdate(bot, update).also {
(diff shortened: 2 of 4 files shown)


telegram/src/main/kotlin/com/github/kotlintelegrambot/dispatcher/handlers/InputChainHandler.kt
```diff
@@ -0,0 +1,65 @@
...
+    val inputChainId: T,
+    val bot: Bot,
+    val update: Update,
+    val message: Message,
+)
+
+class InputChainHandler<T>(
+    private val filter: Filter,
+    private val initialHandler: Handler,
+    private val inputChainIdGen: (previousInputChainId: T?, Update) -> T?,
```
vjgarciag96, line 19: How is this supposed to be used? It's not obvious to me by reading the internal code of library, so I think it will be even more difficult for consumers of the library to understand how to (correctly) use this API. Is it possible to encapsulate this detail in the library?

telegram/src/main/kotlin/com/github/kotlintelegrambot/dispatcher/HandlersDsl.kt
```diff
@@ -165,3 +167,11 @@ fun Dispatcher.dice(body: HandleDice) {
...
     return ExtendedHandler(this, predicate)
 }
+
+fun <T> Handler.inputChain(
+    filter: Filter = Filter.Text,
+    inputChainIdGen: (previousInputChainId: T?, Update) -> T?,
+    handleInputChain: HandleInputChain<T>,
+): Handler {
+    return InputChainHandler(filter, this, inputChainIdGen, handleInputChain)
+}
```
vjgarciag96, line 177: It would be good to also have an example usage of this in the doc and/or one of the sample modules.

telegram/src/main/kotlin/com/github/kotlintelegrambot/dispatcher/handlers/InputChainHandler.kt
```diff
@@ -0,0 +1,65 @@
...
+    private val handleInputChain: HandleInputChain<T>,
+) : Handler {
+    private val chainMap = mutableMapOf<Long, T?>()
```
vjgarciag96, line 22: Also it looks like the value of the map is not really relevant, should this be a hash set of the visited "userId,chatId" instead? That way we can also get rid of `inputChainIdGen`. Or is there another reason why we'd like to have that in the environment object?

telegram/src/main/kotlin/com/github/kotlintelegrambot/dispatcher/handlers/InputChainHandler.kt
```diff
@@ -0,0 +1,65 @@
...
+
+    override fun checkUpdate(update: Update): Boolean =
+        initialHandler.checkUpdate(update) || update.check()
+
+    override suspend fun handleUpdate(bot: Bot, update: Update) {
+        update.message?.let { message ->
+            message.from?.id?.let { userId ->
```
vjgarciag96, line 29: ~~Given that `handleUpdate` is guaranteed to be called after `check` returns `true` we normally make some assumptions (using assertions) about pre-conditions here (for example `message`, `from` and `id` are not null) even if the compiler can't guarantee them~~

~~In the future it might be good to look into wheter we can leverage kotlin contracts to improve this a bit.~~

Nevermind, I realised we can't do that here because `checkUpdate` also checks for initial handler's check which does not guarantee any of these things.

telegram/src/main/kotlin/com/github/kotlintelegrambot/dispatcher/handlers/InputChainHandler.kt
```diff
@@ -0,0 +1,65 @@
...
+                            bot,
+                            update,
+                            message,
+                        ),
+                    )?.let { chainMap[userId] = it }
+                    Unit
+                }
+            }
+        } ?: initialHandler.handleUpdate(bot, update).also {
+            update.takeIf { it.consumed }
```
vjgarciag96, line 44: Why only chaining if the update is consumed? I feel that having to mark the first update as consumed for the chain to work is going to make the API hard to use.

telegram/src/main/kotlin/com/github/kotlintelegrambot/dispatcher/handlers/InputChainHandler.kt
```diff
@@ -0,0 +1,65 @@
...
+
+data class InputChainHandlerEnvironment<T>(
+    val inputChainIdGen: (previousInputChainId: T?, Update) -> T?,
+    val inputChainId: T,
+    val bot: Bot,
+    val update: Update,
+    val message: Message,
+)
+
+class InputChainHandler<T>(
```
vjgarciag96, line 16: I think it would be great to have some tests for this. It would be good as documentation for other maintainers of the library, it would give us more confindence to update this class in the future, it could help us find some edge cases that we may not have considered, etc...

vjgarciag96 (commented):
I like the idea that you are proposing here, just leaving some questions and comments about the implementation.

telegram/src/main/kotlin/com/github/kotlintelegrambot/dispatcher/handlers/InputChainHandler.kt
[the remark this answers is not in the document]
JcMinarro: Good catch!!!
Should we have:
`Map<Long<Map<Long, T?>>` (ChatId, UserId, Identifier)
or:
`Map<String, T?>` ("$chatId$userId, Identifier)

First one should be easier to access.

telegram/src/main/kotlin/com/github/kotlintelegrambot/dispatcher/handlers/InputChainHandler.kt
JcMinarro: I will add some showcases to help other devs to use it and understand how to implement their own chain

telegram/src/main/kotlin/com/github/kotlintelegrambot/dispatcher/handlers/InputChainHandler.kt
JcMinarro: The `T` value is the identifier of the "Input Question".
When the "chain action" is executed, it needs to know "which id it comes from".
Imagine you are creating a form with 3 questions like the one I posted in the PR description, but whenever the question is responded we want to store them on a DB. We need to know which question the user responded, and the value of the `T` is the id used. I used a generic type to allow devs to include their own data type (Maybe more than an int/String value is needed to continue the chain).

telegram/src/main/kotlin/com/github/kotlintelegrambot/dispatcher/handlers/InputChainHandler.kt
JcMinarro: Sure, I will add some tests for this new feature

telegram/src/main/kotlin/com/github/kotlintelegrambot/dispatcher/handlers/InputChainHandler.kt
JcMinarro: I was thinking about it...
I assumed whenever "a question" is processed, they (App devs) should consume it to avoid wrong behavior in other commands that match the "initialHandler" checks (And maybe override the ID stored on this chain), but maybe is a good idea removing this check and rely on the "good implementation" of or users

telegram/src/main/kotlin/com/github/kotlintelegrambot/dispatcher/handlers/InputChainHandler.kt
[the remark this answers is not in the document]
vjgarciag96: First sounds good to me. And if we are always going to access a single identifier for a given (ChatId,UserId) pair another option would be to have our own chain discrimator data class with both that we use as the key (like the second option but without having to concatenate the ids, etc...).

red-avtovo:
@JcMinarro do you find this PR still valid? Do you have any last changes to push? Otherwise I would be happy to accept it

red-avtovo:
Closing due to inactivity

JcMinarro:
@red-avtovo I couldn't complete it yet.
I have a branch that is working on a project I need it, but it only works if your bot is not used from multiple chats because in that case, some collisions could happen as was reported on a previous comment.
I didn't have time to fix it, but I will.
Let me reopen it

red-avtovo:
Oh, I'm glad that the changes are still relevant. Looking forward to reviewing the final changes

red-avtovo:
@JcMinarro just a gentle nudge

seik:
Hi @JcMinarro — thanks for the work and for being upfront about the multi-chat collision issue. Since the implementation isn't ready and the branch has drifted from main, I'm going to close this for now to keep the open-PR list focused.

The `InputChainHandler` idea is still a genuinely useful addition — please feel free to reopen (or open a fresh PR) whenever you have time to wire up the per-chat scoping fix. Really appreciate the original work!

PR #291, opened 2023-06-20
Added a session mechanism. 

ArtdSky:
Added a session mechanism. Session data is kept only in memory

### DIFF
--- a/docs/session.md
+++ b/docs/session.md
@@ -1,0 +1,58 @@
+# Session Management
+
+The SessionManager is a singleton object that manages session data for different chat IDs in a Telegram bot. It provides methods to access, set, get and remove session properties.
+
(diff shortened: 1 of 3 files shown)


vjgarciag96:
Thanks for the contribution @ArtdSky? What's the use case for this API? 

At a first glance I don't see why this should be part of the library rather than being just part of the specific bot app that requires this session mechanism.

ArtdSky:
hello @vjgarciag96!  Use case in dialogs where you want to save the user's answers. In my opinion, it is convenient if the API has the ability to keep some information from the user.  I was inspired by telegraf.js

red-avtovo:
I second @vjgarciag96 on that one, and since there has been no activity for more than a year, I'll close the PR as there is no final decision.
Anyway, I'm grateful for your change proposal @ArtdSky and looking forward to your future PRs

PR #294, opened 2023-07-26
Add `onSuccess()` and `onError()` methods to apply Side Effects to a TelegramBotResult

JcMinarro:
Add `onSuccess()` and `onError()` methods to apply Side Effects to a TelegramBotResult.

It allows us to concatenate the invocation of multiple methods.

```
fun foo(): TelegramBotResult<Int> { ... }

val result = foo()
  .onSuccess {
    // The result was a success and we can apply a side effect with the value
  }
  .onError {
    // The result was an error and we can apply a side effect with the error
  }
```


vjgarciag96 (approved):
Nice! I think we'll need to regenerate the public API files for CI to pass here

PR #306, opened 2023-11-14
Provide 'parse_mode' to sendVideo Telegram API method

thebix:
## Description

Telegram API [sendVideo](https://core.telegram.org/bots/api#sendvideo) method accepts `parse_mode` as a parameter.

Currently, this parameter is not supported by the `sendVideo` method of `kotlin-telegram-bot`
For instance, https://github.com/kotlin-telegram-bot/kotlin-telegram-bot/blob/9fb7d3939d4bc6f273fcc07f316616a33e879383/telegram/src/main/kotlin/com/github/kotlintelegrambot/Bot.kt#L498)

At the same time, other methods like `sendDocument` already [support it](https://github.com/kotlin-telegram-bot/kotlin-telegram-bot/blob/9fb7d3939d4bc6f273fcc07f316616a33e879383/telegram/src/main/kotlin/com/github/kotlintelegrambot/Bot.kt#L476)

This PR adds support for `parse_mode` to all `sendVideo` methods.

## Testing

- [x] Build the library locally and use it as a jar in the test project


telegram/src/main/kotlin/com/github/kotlintelegrambot/Bot.kt
```diff
@@ -502,6 +502,7 @@ class Bot private constructor(
         width: Int? = null,
         height: Int? = null,
         caption: String? = null,
+        parseMode: ParseMode? = null,
```
JcMinarro, line 505: This method is deprecated, we shouldn't update it anymore

JcMinarro (changes requested):

telegram/src/main/kotlin/com/github/kotlintelegrambot/Bot.kt
thebix: Removed `parseMode` from both deprecated methods

JcMinarro (approved):

PR #313, opened 2023-12-26
Add InlineQueryResultsButton

SadeghHayeri:
Hi,
Now we can show web-app URL on top of inline results.
https://core.telegram.org/bots/api#answerinlinequery


seik:
Hi @SadeghHayeri — thanks for the contribution! After 2+ years the branch has conflicts against main and `InlineQueryResultsButton` itself has gained an additional `start_parameter` field on the API side since the PR was opened, so the cleanest path forward is a fresh implementation against current main.

I'm closing this PR but leaving `InlineQueryResultsButton` on the to-do list — happy to take a fresh PR from anyone who wants to pick it up. Really appreciate the original work!

PR #349, opened 2025-04-29
Add support for sending files via InputStream

HeroBrine1st:
This PR adds TelegramFile.ByInputStream, which allows to send files without storing the whole blob on disk or in RAM. Fields of this data class are created independently of [almost same looking Media data class in trixnity](https://gitlab.com/trixnity/trixnity/-/blob/f42affe255ce526b6d3e6f359afbfbdd5997484a/trixnity-clientserverapi/trixnity-clientserverapi-model/src/commonMain/kotlin/net/folivo/trixnity/clientserverapi/model/media/Media.kt) (which has the same purpose), confirming my approach is optimal.  
<sub>I literally realized those two are pretty similar only when started to type this text. Only contentDisposition and filename differ, however contentDisposition is simply a wrapper for filename with metadata</sub>

This PR also adds a sample which closely represents the case we have stumbled upon: copying images from remote server (which requires authorization - not included in the sample) to telegram. There's no point in storing file locally and so this is the clean approach. The sample is clearly not intended as an example usage for library users as it is confusing them, so I expect it to be rejected. However, it is included to a) test it b) show usage (but remember that in our case remote server requires authorization and so TelegramFile.ByUrl will fail).

The changes are fully tested using sample (but only on `sendPhoto`). Filename, contentLength and contentType observed not to be required. There are some concerns which I would like to show using comments on individual lines.

### DIFF
--- a/telegram/src/main/kotlin/com/github/kotlintelegrambot/network/ApiClient.kt
+++ b/telegram/src/main/kotlin/com/github/kotlintelegrambot/network/ApiClient.kt
@@ -17,6 +17,7 @@ import com.github.kotlintelegrambot.entities.TelegramFile
 import com.github.kotlintelegrambot.entities.TelegramFile.ByByteArray
 import com.github.kotlintelegrambot.entities.TelegramFile.ByFile
 import com.github.kotlintelegrambot.entities.TelegramFile.ByFileId
+import com.github.kotlintelegrambot.entities.TelegramFile.ByInputStream
 import com.github.kotlintelegrambot.entities.TelegramFile.ByUrl
 import com.github.kotlintelegrambot.entities.Update
 import com.github.kotlintelegrambot.entities.User
@@ -48,7 +49,9 @@ import com.github.kotlintelegrambot.types.TelegramBotResult
 import com.google.gson.Gson
 import com.google.gson.reflect.TypeToken
 import okhttp3.Interceptor
+import okhttp3.MediaType.Companion.toMediaType
 import okhttp3.MediaType.Companion.toMediaTypeOrNull
+import okhttp3.MultipartBody
 import okhttp3.OkHttpClient
 import okhttp3.RequestBody
 import okhttp3.ResponseBody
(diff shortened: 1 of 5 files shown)


telegram/src/main/kotlin/com/github/kotlintelegrambot/network/ApiClient.kt
```diff
@@ -179,6 +182,20 @@ internal class ApiClient(
             secretToken = secretToken?.toMultipartBodyPart(ApiConstants.SetWebhook.SECRET_TOKEN),
         )
 
+        is ByInputStream -> service.setWebhookWithCertificateAsFile(
+            url = url.toMultipartBodyPart(ApiConstants.SetWebhook.URL),
+            certificate = certificate.let {
+                // set media type if not set (one would probably want assert or override here idk)
+                if(it.contentType == null) it.copy(contentType = MediaTypeConstants.UTF_8_TEXT.toMediaType())
+                else it
```
HeroBrine1st, line 190: I've done this in an attempt to reduce code copying. Assert or override (unconditional copy) are probably better but I don't know

telegram/src/main/kotlin/com/github/kotlintelegrambot/network/Utils.kt
```diff
@@ -27,3 +35,20 @@ fun <T, R> Pair<Response<T?>?, Exception?>.bimap(mapResponse: (T?) -> R, mapErro
...
+    override fun contentLength(): Long = contentLength
+
+    override fun writeTo(sink: BufferedSink) {
+        this@toRequestBody.use {
+            sink.writeAll(it.source())
+        }
+    }
+}
+
+fun TelegramFile.ByInputStream.asMultipartBodyPart(partName: String) = MultipartBody.Part.createFormData(partName, filename ?: partName, stream.toRequestBody(contentType, contentLength))
```
HeroBrine1st, line 53: This is named so because `toMultipartBodyPart` is already used for a receiver on `<T> T`, leading to name conflict without compiler errors but with runtime errors. It's not a concern but it stands out.

UPD: "as" suggests that it is not a transformation but a reuse (e.g. List.reversed vs List.asReversed). Other two providers support copying, while this file provider can't be copied because it holds resource. So, I think it is best to leave as is.

samples/image-send/src/main/kotlin/com/github/kotlintelegrambot/dispatcher/Main.kt
```diff
@@ -0,0 +1,37 @@
...
+fun main() {
+    val httpClient = HttpClient(CIO)
+    val bot = bot {
+        token = "YOUR_API_KEY"
+        timeout = 30
+        logLevel = LogLevel.Network.Headers
+
+        dispatch {
+            command("image") {
+                val response = httpClient.get("https://avatars.githubusercontent.com/u/57418018")
```
HeroBrine1st: <details>
<summary>Btw that's an avatar from this repo</summary>



</details>

samples/image-send/src/main/kotlin/com/github/kotlintelegrambot/dispatcher/Main.kt
```diff
@@ -0,0 +1,37 @@
...
+import io.ktor.client.statement.*
+import io.ktor.http.*
+import io.ktor.utils.io.jvm.javaio.*
+
+fun main() {
+    val httpClient = HttpClient(CIO)
+    val bot = bot {
+        token = "YOUR_API_KEY"
+        timeout = 30
+        logLevel = LogLevel.Network.Headers
```
HeroBrine1st: Also this is imporant enough to be documented somewhere. This __shall__ be a huge frustration source for someone. I've been there and wrote at least two issues (not here) due to this kind of frustration. I've actually stumbled on this exact line already - this is very unobvious!

Setting this to `LogLevel.Network.Body` leads to ProtocolException (if specifying content length) or "no photo present in request" from telegram (if content length is unspecified) because logging consumes the InputStream, leaving actual request with no data to be sent.

HeroBrine1st:
Enough time passed to suppose I should address copilot concerns.
The only valid concern raised by copilot is unused import. There are other concerns I raised myself.

Sample is dropped but available at 8595e9db7f9ef7db210cf636269f444bc4e97f7f, cherry pick for testing (I'll return with UPD for my tests, forgot to do that UPD: still works)

P.s. `./gradlew telegram:apiDump` and even apiCheck do nothing despite new method

seik:
Hi @HeroBrine1st — thanks a lot for the contribution! I've landed your work as 643cb95 on `main` (rebased onto current main, with a small ktlint cleanup of the certificate-MediaType handling in `setWebhookWithCertificateAsFile` and the unused `MultipartBody` import). Authorship preserved on the commit.

I'm opening a follow-up issue to track the remaining coverage gaps (`sendMediaGroup`, `sendVideoNote`, `sendSticker`, `editMessageMedia` — these still don't accept `ByInputStream`) and a missing integration test. Anyone with bandwidth can pick those up.

Thanks again! Closing.

PR #357, opened 2025-12-29
Add RefundedPayment and refundStarPayment fun, minor fixes

mynameisscr:
RefundedPayment is required to be able to process refunds. refundStarPayment is required for a refund. 
Fixed bugs that prevent the library from being built.


telegram/src/main/kotlin/com/github/kotlintelegrambot/entities/User.kt
```diff
@@ -8,7 +8,7 @@ import com.google.gson.annotations.SerializedName
 data class User(
     @SerializedName("id") val id: Long,
     @SerializedName("is_bot") val isBot: Boolean,
-    @SerializedName("is_premium") val isPremium: Boolean?,
+    @SerializedName("is_premium") val isPremium: Boolean? = null,
```
red-avtovo: We can treat it as `false` by default

red-avtovo (commented):
Small comment, otherwise a very solid PR. Thanks a lot for the contribution!

mynameisscr:
Set isPremium false by default and fixed some tests

seik:
Hi @mynameisscr — thanks a lot for this PR! I've landed the refunds part as b28e579 on `main`, with full authorship credit to you. I dropped the User.isPremium / blockquote / test edits since those were already covered by #337 / #347 in the meantime, so the final commit is purely the `RefundedPayment` entity + `refundStarPayment` method + `Message.refundedPayment` field.

Really appreciate the contribution — Stars refunds were a real gap. Closing this PR.

PR #248, opened 2022-06-21
Fix kotlin telegram bot issue #160:

martin-petersen:
- Add possibility to specify the horizontal accuracy of a location

PR #359, opened 2026-01-20
kotlin version up to 2.2.21

useriq:
- up kotlin version to 2.2.21
- fixed isPremium parameters
- fixed tests with coroutines


seik:
Hi @useriq — thanks a lot for tackling this, it was a useful reference for the migration plan. I'm going to handle the bump ourselves and push it straight to main because we want to:

- Go to the latest stable (Kotlin 2.3.21 instead of 2.2.21, coroutines 1.11.0, mockk 1.14.9, gson 2.14.0, etc.).
- Layer it on top of the recent PR landings (#349 InputStream, #357 refunds, #355 textEdit, #248/#247 location, etc.) which have all moved past your branch's base.
- Keep CI on JDK 17 (we updated the workflow yesterday).

Your `DispatcherTest` rewrite for the new coroutines test API was particularly helpful — I'll be reusing the same approach. Closing this PR but you'll see a commit with very similar shape land on main shortly. Really appreciate the work!

PR #360, opened 2026-03-17
Enable concurrent message handling

FlashLight13:
The problem I faced is that suspending inside the handler until the bot receives another message (quite handy for handling callbacks) deadlocks the bot.

The fix is to launch a separated coroutine for each received update. This makes it safe to suspend inside the handler because it will only suspend its dedicated coroutine.

### DIFF
--- a/telegram/src/main/kotlin/com/github/kotlintelegrambot/dispatcher/Dispatcher.kt
+++ b/telegram/src/main/kotlin/com/github/kotlintelegrambot/dispatcher/Dispatcher.kt
@@ -11,8 +11,12 @@ import kotlinx.coroutines.CoroutineDispatcher
 import kotlinx.coroutines.CoroutineScope
 import kotlinx.coroutines.Job
 import kotlinx.coroutines.channels.Channel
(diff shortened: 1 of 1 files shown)


seik:
Hi @FlashLight13 — thanks for the PR and for surfacing the deadlock issue, it's a real one.

I'm going to close this version of the fix though, because as written it changes the dispatcher's default semantics in ways that would break existing bots:

1. `launch` per update means messages from a single chat can now be handled out of order. Telegram guarantees ordered delivery via getUpdates, and most bots (state machines, multi-step commands, conversations) rely on that ordering today.
2. Switching `linkedSetOf` → `CopyOnWriteArraySet` changes the iteration-order semantics for handler dispatch.
3. `update.consume()` and other shared dispatcher state would now race.

Would you be open to a redesign where concurrent handling is opt-in via the bot builder (something like `concurrent = true` or a `coroutineDispatcher` setting that's already there), so existing users keep ordered semantics by default? Happy to take a PR in that shape, or open one ourselves.

Really appreciate you raising this — closing for now in favour of a more conservative approach. Thanks!

